PR #3, opened 2018-11-05
Adding methods from std::function's public interface

rambl30n:


inc/zoo/AnyCallable.h
```diff
@@ -29,6 +29,11 @@ struct AnyCallable;
 /// as they would any type erased value
 template<typename TypeErasureProvider, typename R, typename... Args>
 struct AnyCallable<TypeErasureProvider, R(Args...)>: TypeErasureProvider {
+    static constexpr auto emptyInvoker_ =
```
thecppzoo: I see the need for `emtpyInvoker_` since it is the basis for `empty` and the rest.  At this point there is no reason to keep it as a lambda instead of a function

inc/zoo/AnyCallable.h
```diff
@@ -65,8 +66,60 @@ struct AnyCallable<TypeErasureProvider, R(Args...)>: TypeErasureProvider {
                 const_cast<AnyCallable &>(*this)
             );
     }
+
+    void swap(AnyCallable& other) noexcept {
```
thecppzoo: This is correct, but I prefer to not trust the type erasure provider to give `noexecept` swap, I will change this to `noexcept(noexcept(declval<TypeEraser>().swap(declval<TypeEraser>()))`

inc/zoo/AnyCallable.h
```diff
@@ -65,8 +66,60 @@ struct AnyCallable<TypeErasureProvider, R(Args...)>: TypeErasureProvider {
...
+    const std::type_info& target_type() const noexcept {
+        return empty() ? typeid(void) : this->type();
+    }
+
+    bool empty() const noexcept {
+        return targetInvoker_ == emptyInvoker_;
+    }
+
+    template< class T >
+    T* target() noexcept {
```
thecppzoo: the syntax for pointers and references, `*` and `&` bind to the identifier.  I am in the tiny minority of people who prefer "East pointers and references" over "West pointers and references".  To reduce noise I like declarations of multiple values of the same type, there East pointers and references are better.

inc/zoo/AnyCallable.h
```diff
@@ -65,8 +66,60 @@ struct AnyCallable<TypeErasureProvider, R(Args...)>: TypeErasureProvider {
...
+        using uncvr_t = std::remove_cv_t<std::remove_reference_t<T>>;
```
thecppzoo: `std::decay` does this. Also, I tend to use first-letter upper case camel case for types, except in the cases the components, such as `zoo::function` are so fundamental or similar to the libraries in the official standard that I use the conventions of the standard, but for non-user-facing code, I prefer using my normal capitalization

inc/zoo/AnyCallable.h
```diff
@@ -65,8 +66,60 @@ struct AnyCallable<TypeErasureProvider, R(Args...)>: TypeErasureProvider {
...
 
+// nullptr comparison
+template<typename TypeErasureProvider, typename R, typename... Args>
+bool operator==(AnyCallable<TypeErasureProvider, R(Args...)> const& ac, std::nullptr_t) {
+    return ac.empty();
+}
+
+template<typename TypeErasureProvider, typename R, typename... Args>
+bool operator==(std::nullptr_t, AnyCallable<TypeErasureProvider, R(Args...)> const& ac) {
+    return ac.empty();
```
thecppzoo: This is correct but I prefer chaining, i.e., `return ac == nullptr`

test/AnyCallable.cpp
```diff
@@ -55,4 +55,73 @@ TEST_CASE("function", "[any][type-erasure][functional]") {
...
+        struct MyCallable
+         {
+            auto operator()(int i) {
+                return long(i)*i;
+            }
+        };
+        MyCallable myCallable;
+        SECTION("operator bool()") {
+            zoo::function<long(int)> ac;
+            CHECK(!static_cast<bool>(ac));
```
thecppzoo: This is clear, but I prefer the shorter `bool{ac}`

PR #5, opened 2019-05-19
Adding is_trivially_destructible

rambl30n:


test/inc/zoo/variant.h
```diff
@@ -9,6 +10,25 @@
...
+template<typename... Ts>
+struct Variant;
+
+} // namespace zoo
+
+namespace std {
+
+template<typename... Ts>
+struct is_trivially_destructible<zoo::Variant<Ts...>> :
+    integral_constant<
```
thecppzoo: Nope, specializing `is_trivially_destructible` is not allowed, besides, it does not work either

test/inc/zoo/variant.h
```diff
@@ -73,8 +90,24 @@ template<typename Visitor, typename Var>
...
+
+    template<typename T, typename Enable = void>
+    struct DestroyInvoker {
+    };
+
+    template<template<typename...> class Var, typename... Ts>
+    struct DestroyInvoker<Var<Ts...>,
+            std::enable_if_t<!(std::is_trivially_destructible_v<Ts> && ...)>> {
+        ~DestroyInvoker() {
+            static_cast<Var<Ts...>*>(this)->destroy();
```
thecppzoo: If all this does is to potentially invoke `destroy`, `destroy` could be made a template with `enable_if`

rambl30n: Nope, you need to actually remove the destructor, otherwise a defined destructor (even if it only contains an "if constexpr" that calls destroy will make it non-trivially destructible).

test/inc/zoo/variant.h
rambl30n: Don't know about the magic requirement. The test seems to pass fine.

rambl30n:
The last assertion about the "Derives" example is not true in gcc-9.1. Added a new commit to prove it. Curious to see which compiler are you using.

test/inc/zoo/variant.h
thecppzoo: No disagreement, I meant in the non-trivial case you will have an explicit destructor, in the trivial, an implicit or defaulted destructor, thus this is not necessary

thecppzoo: No disagreement, I meant in the non-trivial case you will have an explicit destructor, in the trivial, an implicit or defaulted destructor, thus this is not necessary

thecppzoo:
It is very interesting GCC accepts the test, Clang doesn’t, not at a computer to show it right now.

test/inc/zoo/variant.h
thecppzoo: I understand your PR better now.  This specialization of `is_trivially_destructible` is not needed, the static assertions succeed because `Variant` has the implicit default destructor, whose triviality is set by its base class `DestroyInvoker` trivial destructibility.  Because of the partial specialization, `DestroyInvoker` is or not trivially destructible correctly.  But still this illegal specialization of `std::is_trivially_destructible` (illegal because it is not one of the exceptions to the rule of not putting something in the `std` namespace) is redundant

test/inc/zoo/variant.h
thecppzoo: This line is "undefined behavior", because this is a base class, the `Variant` instance has already been destroyed, but this wants to use it again.  It is a violation of the abstract machine.  This is a technical thing, but not desirable

thecppzoo:
Can not accept the key idea of destroying in the base class because it is undefined behavior, breaks the abstract machine.  Thanks for the effort

test/inc/zoo/variant.h
thecppzoo: Please check this link:
https://godbolt.org/z/W842oi
You'll see that specializing `is_trivially_destructible` has no effect on derived types

thecppzoo:
I thought Clang was rejecting the code because of this example
https://godbolt.org/z/W842oi
But it is not equivalent to the code here, because `DestroyInvoker` truly is or is not trivially destructible, and this determines the derived classes of `Variant`

rambl30n:
Aawww shucks, you’re totally right. We can’t call derived class methods in
ctor/dtors. I was trying to come up with some clever way to avoid
reimplementing the entire class for the specialization and lost sight of
this basic fact.

Thanks for the feedback.

Julio

PR #11, opened 2022-05-26
Copy in, modify namespace of swar/metalog from pokerbotic.

Scottbruceheart:
Simple copy and namespace fix from pokerbotic.  Haven't attempted to build it in zoo context yet.

### DIFF
--- a/inc/zoo/swar/SWAR.h
+++ b/inc/zoo/swar/SWAR.h
@@ -1,0 +1,179 @@
+#pragma once
+
+/// \file Swar.h SWAR operations
+
+#include "metaLog.h"
+#include <type_traits>
+
+namespace zoo { namespace swar {
+
+/// Repeats the given pattern in the whole of the argument
+/// \tparam T the desired integral type
+/// \tparam Progression how big the pattern is
+/// \tparam Remaining how many more times to copy the pattern
+template<typename T, int Progression, int Remaining> struct BitmaskMaker {
+    constexpr static T repeat(T v) {
+        return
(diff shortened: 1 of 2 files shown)


inc/zoo/swar/SWAR.h
```diff
@@ -0,0 +1,179 @@
+#pragma once
+
+/// \file Swar.h SWAR operations
+
+#include "metaLog.h"
```
thecppzoo, line 5: I think relative inclusion paths are counterproductive: I like the ability to have parts of the inclusion tree in different file paths and let the compilation merge them into a single one. 
Change to include `zoo/swar/metaLog.h` or perhaps move this to the top level as a top-level utility

thecppzoo: Also, I have this custom:
1. Include first the most specific headers, in this case, the ones in the `zoo` library.
2. For each section, leave a space.  In this case a blank line separating the inclusion of metaLog from type-traits

inc/zoo/swar/SWAR.h
```diff
@@ -0,0 +1,179 @@
...
+}
+
+template<int Level>
+constexpr uint64_t popcount(uint64_t a) {
+    return detail::Selector_impl<2 < Level>::template execute<Level>(a);
+}
+
+static_assert(0x210 == popcount<0>(0x320), "");
+static_assert(0x4321 == popcount<1>(0xF754), "");
+static_assert(0x50004 == popcount<3>(0x3E001122), "");
```
thecppzoo, line 105: These `static_assert` are not very helpful, let us segregate them to a test file (a `swar.cpp` file, perhaps)

PR #13, opened 2022-05-28
Add pokerbotic swar to zoo.

Scottbruceheart:
Copies over swar/metalog from pokerbotic, adds a shim test, add cmakelists entry for swar testing.

### DIFF
--- a/inc/zoo/swar/SWAR.h
+++ b/inc/zoo/swar/SWAR.h
@@ -1,0 +1,180 @@
+#pragma once
+
+/// \file Swar.h SWAR operations
+
+#include "zoo/swar/metaLog.h"
+
+#include <type_traits>
+
+namespace zoo { namespace swar {
+
+/// Repeats the given pattern in the whole of the argument
+/// \tparam T the desired integral type
+/// \tparam Progression how big the pattern is
+/// \tparam Remaining how many more times to copy the pattern
+template<typename T, int Progression, int Remaining> struct BitmaskMaker {
+    constexpr static T repeat(T v) {
+        return
+            BitmaskMaker<T, Progression, Remaining - 1>::repeat(v | (v << Progression));
+    }
+};
+template<typename T, int P> struct BitmaskMaker<T, P, 0> {
+    constexpr static T repeat(T v) { return v; }
+};
+
+/// Front end to \c BitmaskMaker with the repeating count set to the whole size
+template<int size, typename T>
+constexpr T makeBitmask(T v) {
+    return BitmaskMaker<T, size, sizeof(T)*8/size>::repeat(v);
+}
+
+/// Core implementation details
+namespace detail {
+    template<int level>
+    constexpr auto popcountMask =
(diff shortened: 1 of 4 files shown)


test/swar/BasicOperations.cpp
```diff
@@ -0,0 +1,19 @@
...
+using namespace zoo;
+
+TEST_CASE(
+    "Basic SWAR",
+    "[swar]"
+) {
+    uint8_t v = 0x3;
+    SECTION("Simple Mask") {
+      auto mask = swar::makeBitmask<8, uint8_t>(v);
+      CHECK(mask == 0x08080808ull);
```
thecppzoo, line 17: In the interest of brevity, perhaps we should make a 32 bit or perhaps a 16 bit test instead, of nibbles?

inc/zoo/swar/SWAR.h
```diff
@@ -0,0 +1,180 @@
...
+}
+
+static_assert(0x210 == popcount<0>(0x320), "");
+static_assert(0x4321 == popcount<1>(0xF754), "");
+static_assert(0x50004 == popcount<3>(0x3E001122), "");
+
+
+template<typename T> constexpr typename std::make_unsigned<T>::type msb(T v) {
+    return 8*sizeof(T) - 1 - __builtin_clzll(v);
+}
```
thecppzoo, line 111: This should be called something like `bitIndex`, `msb` invites to think about the msb...

inc/zoo/swar/SWAR.h
```diff
@@ -0,0 +1,180 @@
...
+    constexpr auto msbMask = makeBitmask<Size, T>(msb);
+    constexpr auto subtraend = makeBitmask<Size, T>(N);
+    auto adjusted = v.value() | msbMask;
+    auto rv = adjusted - subtraend;
+    rv &= msbMask;
+    return rv;
+}
+
+static_assert(
+    0x80880008 == greaterEqualSWAR<3>(SWAR<4, uint32_t>(0x32451027)).value(),
```
thecppzoo, line 176: These tests... let's improve this afterwards

test/swar/BasicOperations.cpp
```diff
@@ -0,0 +1,19 @@
...
+TEST_CASE(
+    "Basic SWAR",
+    "[swar]"
+) {
+    uint8_t v = 0x8;
+    SECTION("Simple Mask") {
+      auto mask = swar::makeBitmask<8, uint32_t>(v);
+      CHECK(mask == 0x08080808ull);
+    }
+}
```
thecppzoo, line 19: Test the things taht can be tested at compilation time at compilation time.  As of now, there's no need to include catch here either

thecppzoo (approved):
I'd prefer you make tests at compilation time

inc/zoo/swar/SWAR.h
Scottbruceheart: I certainly thought msb meant 'most significant bit', msb_index does seem like a more clear name.

PR #14, opened 2022-06-05
[SWAR] Add various SWAR ops, support ops, and tests.

Scottbruceheart:
Improve SWAR test.
Add horizontal equality, remove operator== on SWAR.
Add broadcast.
Aadd (isolate,clear)x(least significant bit, least significant bits)
Remove most non-explicit conversions.
Add MSB indexes as constexpr for boolean swar. 
Add const/noecept enthusiastically.
Add u64 lsb_index.  
Add greaterequal_msb_off. 
Add basic popcount static_asserts.


inc/zoo/swar/SWAR.h
```diff
@@ -8,23 +8,28 @@
...
+using u16 = uint16_t;
+using u8 = uint8_t;
+
 /// Repeats the given pattern in the whole of the argument
 /// \tparam T the desired integral type
 /// \tparam Progression how big the pattern is
 /// \tparam Remaining how many more times to copy the pattern
 template<typename T, int Progression, int Remaining> struct BitmaskMaker {
-    constexpr static T repeat(T v) {
+    constexpr static T repeat(T v) noexcept {
```
thecppzoo, line 21: Thanks for making this "`noexcept`-correct"  John Lakos would say this is ridiculous but out of experience I know it is not, the noexceptness will eventually allow us to have guaranteed-success operations and those that are not.  One example: invoking "count leading zeroes" (binary logarithm) or trailing zeroes (index within the SWAR) are not guaranteed to work if the number is 0, this makes them not-safe.  We can make an algebraic system in which these things are automatically selected, for much better correctness and perhaps performance.

inc/zoo/swar/SWAR.h
```diff
@@ -83,99 +88,153 @@ namespace detail {
...
+    constexpr explicit operator T() const noexcept { return m_v; }
 
     constexpr T value() const noexcept { return m_v; }
 
-    constexpr SWAR operator|(SWAR o) { return SWAR(m_v | o.m_v); }
-    constexpr SWAR operator&(SWAR o) { return SWAR(m_v & o.m_v); }
-    constexpr SWAR operator^(SWAR o) { return SWAR(m_v ^ o.m_v); }
+    constexpr SWAR operator|(SWAR o) const noexcept { return SWAR(m_v | o.m_v); }
+    constexpr SWAR operator&(SWAR o) const noexcept { return SWAR(m_v & o.m_v); }
+    constexpr SWAR operator^(SWAR o) const noexcept { return SWAR(m_v ^ o.m_v); }
```
thecppzoo, line 127: Damn design of C++ that forces you to do these busy-work things.

I believe this is better with an X-List macro, do you disagree?

inc/zoo/swar/SWAR.h
```diff
@@ -83,99 +88,153 @@ namespace detail {
...
 
-    constexpr bool operator==(SWAR o) { return m_v == o.m_v; }
-
-    constexpr T at(int position) {
-        constexpr auto filter = (T(1) << Size) - 1;
-        return filter & (m_v >> (Size * position));
+    constexpr T at(int position) const noexcept {
+        constexpr auto filter = (T(1) << NBits) - 1;
+        return filter & (m_v >> (NBits * position));
```
thecppzoo, line 131: I believe we will run into issues if we keep using `NBits * position` because when it is a multiplication, it is painfully expensive.  We ought to create the cheapo version as a partial template specialization for the case of powers of 2

inc/zoo/swar/SWAR.h
```diff
@@ -83,99 +88,153 @@ namespace detail {
...
         auto mask = ~invertedMask;
         return SWAR(m_v & mask);
     }
 
-    constexpr int top() { return msb_index(m_v) / Size; }
-    constexpr int fastIndex() { return __builtin_ctzll(m_v) / Size; }
+    /// Notably, this is the SWAR lane index that contains the MSB.
+    constexpr int top() const noexcept { return msb_index(m_v) / NBits; }
+    /// Notably, this is the SWAR lane index that contains the lsb.
+    constexpr int lsbIndex() const noexcept { return __builtin_ctzll(m_v) / NBits; }
```
thecppzoo, line 144: What is the meaning of "Notably" here?

thecppzoo: I might adopt your implicit suggestion of calling SWAR elements "lanes"

inc/zoo/swar/SWAR.h
```diff
@@ -83,99 +88,153 @@ namespace detail {
...
+
+template<typename T = uint64_t>
+constexpr auto isolateLSB(T v) {
+  return v & ~clearLSB(v);
+}
+
+template<int NBits, typename T = uint64_t>
+constexpr auto maskLowBits() {
+  return (T(1)<<(NBits-1)) | ((T(1)<<(NBits-1))-1);
+}
```
thecppzoo, line 176: This ought to be an inline variable,
```c++
temp....
constexpr auto T LowBitsMask = ...
```

thecppzoo: This looks incorrect: `1 << (NBits - 1)` or'd with itself minus one is equal to setting all bits

inc/zoo/swar/SWAR.h
Scottbruceheart: noexcept as a concept is very powerful. I agree it is nice to set us up for a regime in which we can set up always most-correct behavior.

inc/zoo/swar/SWAR.h
Scottbruceheart: It is better that way, I just haven't changed it yet.

inc/zoo/swar/SWAR.h
Scottbruceheart: I'm not sure we're actually using at anywhere yet, but I do agree it is too expensive.  I was planning to ignore the problem until I thought of something smarter.

inc/zoo/swar/SWAR.h
Scottbruceheart: as opposed to msb_index/lsb_index that find those indexes against a u64 or other integral type.

inc/zoo/swar/SWAR.h
```diff
@@ -83,99 +88,153 @@ namespace detail {
...
+
+template<int NBits, typename T = uint64_t>
+constexpr auto clearLSBits(T v) {
+  return v &(~(maskLowBits<NBits>() << metaLogFloor(isolateLSB<T>(v))));
+}
+
+template<int NBits, typename T = uint64_t>
+constexpr auto isolateLSBits(T v) {
+  return v &(maskLowBits<NBits>() << metaLogFloor(isolateLSB<T>(v)));
+}
```
thecppzoo, line 186: Oh, no, these are unnecessarily expensive:
`isolateLSBs`:
`v & ~(v - 1)`
`v-1` has all MSBs equal to `v`, the rest, including the LSB set, are different to `v`, hence its bitwise-negation is equal to `v` only in the LSB

inc/zoo/swar/SWAR.h
```diff
@@ -83,99 +88,153 @@ namespace detail {
...
+
+template<int NBits, typename T = uint64_t>
+constexpr auto broadcast(SWAR<NBits, T> v) {
+  constexpr T Ones = makeBitmask<NBits, T>(SWAR<NBits, T>{1}.value());
+  return SWAR<NBits, T>(T(v) * Ones);
```
thecppzoo, line 191: Please make a note this assumes the input has **only the desired value in the least significant "lane"**

inc/zoo/swar/SWAR.h
```diff
@@ -83,99 +88,153 @@ namespace detail {
...
+}
+
+template<int NBits, typename T>
+struct BooleanSWAR: SWAR<NBits, T> {
+    // Booleanness is stored in MSB of a given swar.
```
thecppzoo, line 196: Unclear comment.
Please attempt to explain that the MSBs of each element/"lane" contains the boolean, other bits are zero

inc/zoo/swar/SWAR.h
```diff
@@ -83,99 +88,153 @@ namespace detail {
...
+    constexpr explicit BooleanSWAR(T v): SWAR<NBits, T>(v) {}
 
-    constexpr BooleanSWAR clear(int bit) {
-        constexpr auto Bit = T(1) << (Size - 1);
-        return this->m_v ^ (Bit << (Size * bit)); }
-    constexpr int best() { return this->top(); }
+    constexpr BooleanSWAR clear(int bit) const noexcept {
+        constexpr auto Bit = T(1) << (NBits - 1);
+        return this->m_v ^ (Bit << (NBits * bit)); }
+    constexpr int best() const noexcept { return this->top(); }
```
thecppzoo, line 203: Hmmm... this name is not useful.  Mind you, I made a lot of mistakes like these, thanks for going over them

inc/zoo/swar/SWAR.h
```diff
@@ -83,99 +88,153 @@ namespace detail {
...
+
+    constexpr operator bool() const noexcept { return this->m_v; }
```
thecppzoo, line 205: Remove.  Rename.  Make it an explicit function like `verticalBool`?

inc/zoo/swar/SWAR.h
```diff
@@ -83,99 +88,153 @@ namespace detail {
...
+
+    constexpr BooleanSWAR operator not() const noexcept {
+        return MSBs ^ *this;
+    }
```
thecppzoo, line 209: Please make a note this is not `compl` or `~` to avoid inadvertent confusion

inc/zoo/swar/SWAR.h
```diff
@@ -83,99 +88,153 @@ namespace detail {
...
 };
 
-template<int N, int Size, typename T>
-constexpr BooleanSWAR<Size, T> greaterEqualSWAR(SWAR<Size, T> v) {
-    static_assert(1 < Size, "Degenerated SWAR");
-    static_assert(metaLogCeiling(N) < Size, "N is too big for this technique");
-    constexpr auto msbPos  = Size - 1;
+template<int NBits, typename T>
+constexpr BooleanSWAR<NBits, T> greaterEqual_MSB_off(SWAR<NBits, T> left, SWAR<NBits, T> right) noexcept {
+    constexpr auto MSOnes = BooleanSWAR<NBits, T>::MSBs;
```
thecppzoo, line 220: ` MSOnes` invites the confusion witht he register of the numer 1 in each "lane".  `MSBs`?

inc/zoo/swar/SWAR.h
```diff
@@ -83,99 +88,153 @@ namespace detail {
...
+    return SWAR<NBits, T>{
+        (((left.value() | MSOnes) - right.value()) & MSOnes)
+    };
```
thecppzoo, line 223: the `values()` projection is not needed, since you should be able to rely on the overloaded operators for `SWAR`

inc/zoo/swar/SWAR.h
```diff
@@ -83,99 +88,153 @@ namespace detail {
...
+        return this->m_v ^ (Bit << (NBits * bit)); }
+    constexpr int best() const noexcept { return this->top(); }
+
+    constexpr operator bool() const noexcept { return this->m_v; }
+
+    constexpr BooleanSWAR operator not() const noexcept {
+        return MSBs ^ *this;
+    }
+ private:
+    constexpr BooleanSWAR(SWAR<NBits, T> initializer) noexcept: SWAR<NBits, T>(initializer) {}
```
thecppzoo, line 211: I think this is a good idea, of keeping the constructors private so that only clear ways to make a boolean SWAR exist

test/swar/BasicOperations.cpp
```diff
@@ -6,15 +6,146 @@
 #include <type_traits>
 
 using namespace zoo;
+using namespace zoo::swar;
+
+#define HE(nbits, t, v0, v1) static_assert(horizontalEquality<nbits, t>(SWAR<nbits, t>(v0), SWAR<nbits, t>(makeBitmask<nbits, t>(v1))));
+HE(8, u64, 0x0808'0808'0808'0808, 0x8);
```
thecppzoo, line 12: Why is the last element a bit-mask?
This would be a chance to use octal notation for 3 bits

test/swar/BasicOperations.cpp
```diff
@@ -6,15 +6,146 @@
...
+static_assert(0xFC == clearLSBits<2, u8>(0xFF));
+
+static_assert(0x80 == isolateLSBits<2, u8>(0x80));
+static_assert(0xC0 == isolateLSBits<2, u8>(0xC0));
+static_assert(0x60 == isolateLSBits<2, u8>(0xE0));
+static_assert(0x30 == isolateLSBits<2, u8>(0xF0));
+static_assert(0x18 == isolateLSBits<2, u8>(0xF8));
+static_assert(0x03 == isolateLSBits<2, u8>(0xFB));
+static_assert(0x0C == isolateLSBits<2, u8>(0xFC));
+static_assert(0x03 == isolateLSBits<2, u8>(0xFF));
```
thecppzoo, line 110: I think the suffix `Bits` does not add over `Bs`

thecppzoo: Sorry, but I don't like using the element size of 2 for these tests, the convenient way to do this, I htik is to use nibbles, 4 bits.

test/swar/BasicOperations.cpp
```diff
@@ -6,15 +6,146 @@
...
+static_assert(0x0000'0008 == greaterEqual_MSB_off<4, u32>(SWAR<4, u32>(0x0123'4567), SWAR<4, u32>(0x7777'7777)).value());
+
+static_assert(0x8000'0000 == greaterEqual_MSB_off<4, u32>(SWAR<4, u32>(0x0000'0000), SWAR<4, u32>(0x0123'4567)).value());
+static_assert(0x8800'0000 == greaterEqual_MSB_off<4, u32>(SWAR<4, u32>(0x1111'1111), SWAR<4, u32>(0x0123'4567)).value());
+static_assert(0x8880'0000 == greaterEqual_MSB_off<4, u32>(SWAR<4, u32>(0x2222'2222), SWAR<4, u32>(0x0123'4567)).value());
+static_assert(0x8888'0000 == greaterEqual_MSB_off<4, u32>(SWAR<4, u32>(0x3333'3333), SWAR<4, u32>(0x0123'4567)).value());
+static_assert(0x8888'8000 == greaterEqual_MSB_off<4, u32>(SWAR<4, u32>(0x4444'4444), SWAR<4, u32>(0x0123'4567)).value());
+static_assert(0x8888'8800 == greaterEqual_MSB_off<4, u32>(SWAR<4, u32>(0x5555'5555), SWAR<4, u32>(0x0123'4567)).value());
+static_assert(0x8888'8880 == greaterEqual_MSB_off<4, u32>(SWAR<4, u32>(0x6666'6666), SWAR<4, u32>(0x0123'4567)).value());
+static_assert(0x8888'8888 == greaterEqual_MSB_off<4, u32>(SWAR<4, u32>(0x7777'7777), SWAR<4, u32>(0x0123'4567)).value());
```
thecppzoo, line 151: These are not clear tests, becau7se it is super hard to check humanly.
Perhaps:
```c++
s_a(greaterEqual....
0x4444'4444,
0x0123'4567) ==
0x8888'8000
``` 
^^^ the user has a spatial clue to compare

thecppzoo (changes requested):
I'm super grateful to you for this work, please make the changes requested and we'll merge

inc/zoo/swar/SWAR.h
thecppzoo: What I mean is we will provide the API for the powers of 2 so that the user unambiguously knows it is using the fast and to prevent the mistake when not.  However, we ought to keep `at` in the way we have it now for generality

inc/zoo/swar/SWAR.h
thecppzoo: Still not agreeing with "notably"

inc/zoo/swar/SWAR.h
Scottbruceheart: agreed

inc/zoo/swar/SWAR.h
Scottbruceheart: I'm fine with removing and not replacing until we actually need it.

inc/zoo/swar/SWAR.h
Scottbruceheart: MSBitsOn maybe?  That is the mask of 'which bits per SWAR or SWAR lane are treated as semantically boolean', but I don't know how to say that succinctly. MSBMask or maskMSB?

inc/zoo/swar/SWAR.h
Scottbruceheart: I must have removed those operators, or am catching an implicit conversion somewhere I missed. Let me check ...

inc/zoo/swar/SWAR.h
```diff
@@ -83,99 +88,153 @@ namespace detail {
...
-    constexpr BooleanSWAR(T v): SWAR<Size, T>(v) {}
+/// Defining operator== on base SWAR types is entirely too error prone. Force a verbose invocation.
+template<int NBits, typename T = uint64_t> constexpr auto horizontalEquality(SWAR<NBits, T> left, SWAR<NBits, T> right) {
+  return left.m_v == right.m_v;
+}
+
+template<int NBits, typename T = uint64_t>
+constexpr auto isolate(T pattern) {
+  return pattern & ((T(1)<<NBits)-1);
+}
```
thecppzoo, line 166: I've just discovered a serious error: imagine the element size is equal to swar size (this ought to be fully supported, BTW), then `1 << NBits` is either a no-op and then the code is consuming whatever garbage was there, or perhaps 0.  This case is not even *defined*.  We somehow have to make sure we **never** shift as many bits as the count of bits in the integer type.

inc/zoo/swar/SWAR.h
```diff
@@ -107,10 +107,12 @@ constexpr uint64_t popcount(uint64_t a) noexcept {
 }
 
 
+/// Index into the bits of the type T that contains the MSB.
 template<typename T> constexpr typename std::make_unsigned<T>::type msb_index(T v) noexcept {
     return 8*sizeof(T) - 1 - __builtin_clzll(v);
 }
 
+/// Index into the bits of the type T that contains the LSB.
 template<typename T> constexpr typename std::make_unsigned<T>::type lsb_index(T v) noexcept {
```
thecppzoo, line 116: why the `snake_case` instead of the `bactrianCase`?

inc/zoo/swar/SWAR.h
Scottbruceheart: Agreed.

inc/zoo/swar/SWAR.h
```diff
@@ -138,9 +140,12 @@ template<int NBits, typename T = uint64_t> struct SWAR {
...
     }
 
-    /// Notably, this is the SWAR lane index that contains the MSB.
+    /// The SWAR lane index that contains the MSB.  It is not the bit index of
+    /// the underlying type that contains the MSB.
+    /// IE: 8 bit wide 32 bit SWAR: 0x0040'0000 will return 5, not 43.
     constexpr int top() const noexcept { return msb_index(m_v) / NBits; }
-    /// Notably, this is the SWAR lane index that contains the lsb.
+    /// The SWAR lane index that contains the LSB.  It is not the bit index of
+    /// the underlying type that contains the LSB.
```
thecppzoo, line 148: 1. I prefer to never copy and paste, this extends to comments.
2. Alas, the comment is misleading, since here are two possible understandings: index in bits or in elements ("lanes"), the trailing part of "of the underlying type...." is unnecessary and its presence is confusing

test/swar/BasicOperations.cpp
Scottbruceheart: It just happened that way (in that I wanted HE to test something function based against something constant, and make bitmask was it.)

inc/zoo/swar/SWAR.h
```diff
@@ -160,29 +165,37 @@ constexpr auto isolate(T pattern) {
   return pattern & ((T(1)<<NBits)-1);
 }
 
+/// Clears the lowest bit set in type T
 template<typename T = uint64_t>
 constexpr auto clearLSB(T v) {
   return v & (v - 1);
 }
 
+// Sets only the loest bit set in type T
```
thecppzoo, line 174: Confusing comment, perhaps "leaves ON only the lowest bit set, or ~0 (all ones) for a zero input"

inc/zoo/swar/SWAR.h
Scottbruceheart: Yeah, I definitely noticed that edge case. Not exactly sure what to do about it.

inc/zoo/swar/SWAR.h
thecppzoo: It is confusing because it does not "set" a bit that was off (unless the input is zero)

inc/zoo/swar/SWAR.h
Scottbruceheart: It was there before, fixing.

inc/zoo/swar/SWAR.h
Scottbruceheart: LOEST hahahaha

test/swar/BasicOperations.cpp
Scottbruceheart: I wanted to use minimal size testing. Truly I should add 2,3,4,etc bits to various sizes to be truly paranoid in testing.

inc/zoo/swar/SWAR.h
Scottbruceheart: operator- isn't on swar, will add next PR (well, adjust to xlist and add operator-), leaving this for now.

inc/zoo/swar/SWAR.h
Scottbruceheart: Some other writer around SWAR/SIMD calls them lanes, but I can't remember where I first saw it.  Seems like a useful mental model.

inc/zoo/swar/SWAR.h
Scottbruceheart: Done

inc/zoo/swar/SWAR.h
Scottbruceheart: I thought the abstraction was that we'd interpret whatever happened to be in the MSB of each lane as a boolean, regardless of what the remaining bits are?  Basically we'd need to mask it with maskLaneMSB before treating it as a boolean.

inc/zoo/swar/SWAR.h
```diff
@@ -117,6 +117,11 @@ template<typename T> constexpr typename std::make_unsigned<T>::type lsbIndex(T v
     return __builtin_ctzll(v) + 1;
 }
 
+/// Core abstraction around SIMD Within A Register (SWAR).  Specifies 'lanes'
+/// of NBits width against a type T, and provides an abstraction for performing
+/// SIMD operations against that primitive type T treated as a SIMD register.
+/// SWAR operations are usually constant time, log(lane count) cost, or O(lane count) cost.
+/// Certain computaitonal workloads can be materially sped up using SWAR techniques.
```
thecppzoo, line 124: typo

inc/zoo/swar/SWAR.h
```diff
@@ -196,6 +201,8 @@ constexpr auto isolateLSBits(T v) {
   return v &(lowMask << metaLogFloor(isolateLSB<T>(v)));
```
thecppzoo, line 201: If you use triple slashes, that's a Doxygen comment format, make sure you intend to pass this to Doxygen, otherwise, better don't bother because tools like XCode parse them and would be confused

inc/zoo/swar/SWAR.h
```diff
@@ -83,99 +88,173 @@ namespace detail {
...
 
-template<int Size, typename T>
-struct BooleanSWAR: SWAR<Size, T> {
-    constexpr BooleanSWAR(T v): SWAR<Size, T>(v) {}
+/// Defining operator== on base SWAR types is entirely too error prone. Force a verbose invocation.
+template<int NBits, typename T = uint64_t> constexpr auto horizontalEquality(SWAR<NBits, T> left, SWAR<NBits, T> right) {
+  return left.m_v == right.m_v;
+}
+
+/// Isolating more than bits in type currently results in disaster.
```
thecppzoo, line 165: `NBits`

inc/zoo/swar/SWAR.h
thecppzoo: Using "lanes" the clarity  improved a lot

inc/zoo/swar/SWAR.h
thecppzoo: For things to work best, the "lanes" should have only the MSB set when "true", all false otherwise.
Because of this property, all the operations that generate a BooleanSWAR must be very carefully thought so that they guarantee that property without non-essential performance penalties.  Let us discourage the conversion from SWAR or integer to BooleanSWAR so that we don't do it "manually" but systematically identify when/how we generate an array of bools

test/swar/BasicOperations.cpp
```diff
@@ -6,15 +6,198 @@
...
+            0x0000'8888)
+GE_MSB_TEST(0x0123'4567,
+            0x5555'5555,
+            0x0000'0888)
+GE_MSB_TEST(0x0123'4567,
+            0x6666'6666,
+            0x0000'0088)
+GE_MSB_TEST(0x0123'4567,
+            0x7777'7777,
+            0x0000'0008)
```
thecppzoo, line 178: Thanks for the attempt to be comprehensive, but alas, this is not!:
We ought to look for a "problem" of a "lane" that might "borrow" from the adjoining lane, or failing to compare against zero, or something like that.
Also, tests that are entirely redundant actually reduce clarity.  Please identify the cases you are testing, whether the tests are comprehensive, and do not introduce them redundantly

inc/zoo/swar/SWAR.h
```diff
@@ -83,99 +88,173 @@ namespace detail {
...
+        constexpr auto filter = (T(1) << NBits) - 1;
+        auto invertedMask = filter << (NBits * position);
         auto mask = ~invertedMask;
         return SWAR(m_v & mask);
     }
 
-    constexpr int top() { return msb_index(m_v) / Size; }
-    constexpr int fastIndex() { return __builtin_ctzll(m_v) / Size; }
+    /// The SWAR lane index that contains the MSB.  It is not the bit index of the MSB.
+    /// IE: 8 bit wide 32 bit SWAR: 0x0040'0000 will return 5, not 43.
```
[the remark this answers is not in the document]
Scottbruceheart: Yeah, I definitely was fitting more than 4 bits per ascii symbol, and yet not enough to reach 32. Good catch.

inc/zoo/swar/SWAR.h
Scottbruceheart: It seemed like you wanted to have comments of this form. Do we not doxygen /autoparse comments?

inc/zoo/swar/SWAR.h
Scottbruceheart: Good catch.

Scottbruceheart: I do not understand why attempting to use ~ results in "can't shift negative numbers left" with unsigned integral types.

test/swar/BasicOperations.cpp
Scottbruceheart: Is fine, if you don't like redundancy, easy to cut down.  Very easy to generate, also.
These are all testing MSB off, so I'm not sure what edge case is interesting?  Lanes should be isolated, but shifted tests some.

thecppzoo (approved):
A few things to modify but let's merge

PR #15, opened 2022-08-06
Draft of SWAR binary logarithm

thecppzoo:
Changes:
1. Introduces calculation of the logarithm on a per-SWAR element basis
2. Updates the code to prevent known deficiencies from Pokerbotic:
    1. It is an invitation for problems to have the most primitive way to make bitmasks from a given pattern a `constexpr` function because it may be invoked from the runtime resulting in a log(N) execution time instead of constant.  Instead of a function it is now a template that can only be used from compilation time.
        1. I decided to keep this primitive mechanism instead of the "broadcasting" because it can be scaled up to any size, `broadcast` relies on multiplication of the size of the register, which for SIMD registers won't be available.
3. For similar reasons the functions to calculate popcounts ("Hamming weights") are now "structs" with a `constexpr` function
4. The files were rearranged so that we now have
    1. Non-SWAR functions
    2. The SWAR functions are implemented in terms of the non-SWAR


inc/zoo/meta/BitmaskMaker.h
```diff
@@ -0,0 +1,48 @@
...
+        >::value;
+};
+template<typename T, T Current, int CurrentSize>
+struct BitmaskMaker_impl<T, Current, CurrentSize, false> {
+    constexpr static T value = Current;
+};
+
+} // detail
+
+/// Repeats the given pattern in the whole of the argument
```
Scottbruceheart, line 31: Suggest providing example usage in comment.

inc/zoo/meta/BitmaskMaker.h
```diff
@@ -0,0 +1,48 @@
...
+/// \tparam CurrentSize the number of bits in the pattern
+template<typename T, T Current, int CurrentSize>
+struct BitmaskMaker {
+    constexpr static auto value =
+        detail::BitmaskMaker_impl<
+            T, Current, CurrentSize,  CurrentSize*2 <= sizeof(T)*8
+        >::value;
+};
+
+static_assert(0xF0F0 == BitmaskMaker<uint16_t, 0xF0, 8>::value);
```
Scottbruceheart, line 43: static asserts are only multiples of two: is there a restriction?
`BitmaskMaker(uint16_t, 0xC5, 5>::value == ?`

thecppzoo: It seems you mean a power of two, the very next line has an example of a pattern of 12 bits

PR #17, opened 2022-08-10
Add SWARWithSublanes, SWAR w/2 arbitrary width 'sublanes'.

Scottbruceheart:
Add SWARWithSublanes, a useful abstraction for having two arbitrary width sublanes associated with a given SWAR.

Also switches usage of high/low and similar to most/least for clarity.

### DIFF
--- a/inc/zoo/swar/SWAR.h
+++ b/inc/zoo/swar/SWAR.h
@@ -144,6 +144,16 @@ template<int NBitsMost, int NBitsLeast, typename T = uint64_t> struct SWARWithSu
         return most(pos) >> (LaneBits*pos)>> NBitsLeast;
     }
 
+    // Blits most sig bits into least significant bits. Experimental.
+    constexpr auto flattenMostToLeast(u32 pos) const noexcept {
+        return (this->m_v >> NBitsLeast) & LeastMask;
+    }
+
+    // Blits least sig bits into most significant bits. Experimental.
+    constexpr auto promoteLeastToMost(u32 pos) const noexcept {
+        return (this->m_v << NBitsMost) & MostMask;
+    }
+
     // Sets the lsb sublane at |pos| with least significant NBitsLeast of |in|
     constexpr auto least(T in, int pos) const noexcept {
         constexpr auto filter = (T(1) << LaneBits) - 1;
--- a/inc/zoo/swar/SWAR.h
+++ b/inc/zoo/swar/SWAR.h
@@ -40,15 +40,19 @@ template<typename T> constexpr typename std::make_unsigned<T>::type lsbIndex(T v
 /// SWAR operations are usually constant time, log(lane count) cost, or O(lane count) cost.
 /// Certain computational workloads can be materially sped up using SWAR techniques.
 template<int NBits, typename T = uint64_t> struct SWAR {
+    static constexpr inline auto LaneCount = sizeof(T) * 8 / NBits;
     SWAR() = default;
     constexpr explicit SWAR(T v): m_v(v) {}
     constexpr explicit operator T() const noexcept { return m_v; }
 
     constexpr T value() const noexcept { return m_v; }
 
-    constexpr SWAR operator|(SWAR o) const noexcept { return SWAR(m_v | o.m_v); }
-    constexpr SWAR operator&(SWAR o) const noexcept { return SWAR(m_v & o.m_v); }
-    constexpr SWAR operator^(SWAR o) const noexcept { return SWAR(m_v ^ o.m_v); }
+    constexpr SWAR operator~() const noexcept { return SWAR{~m_v}; }
+    #define SWAR_BINARY_OPERATORS_X_LIST \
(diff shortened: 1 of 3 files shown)


inc/zoo/swar/SWAR.h
```diff
@@ -157,57 +161,152 @@ template<int NBits, typename T = uint64_t> struct SWAR {
     T m_v;
 };
 
+// SWAR is a useful abstraction for performing computations in lanes overlaid
+// over any given integral type.
+// Doing additions, subtractions, and compares via SWAR techniques requires an
+// extra bit per lane be available past the lane size, _or_ knowledge that both
+// of your MSBs are set 0 (leaving space for the operation).  Similarly, doing
+// multiplications via SWAR techniques require double bits per lane (unless you
+// can bound your inputs at half lane size).
```
thecppzoo, line 170: sed/bound/bind/

inc/zoo/swar/SWAR.h
```diff
@@ -157,57 +161,152 @@ template<int NBits, typename T = uint64_t> struct SWAR {
...
+        return this->m_v & keep;
+    }
+
+    // Returns only the most significant bits at specified position, 'decoded' to their integer value.
+    constexpr auto mostFlat(u32 pos) const noexcept {
+        return most(pos) >> (LaneBits*pos)>> NBitsLeast;
+    }
+
+    // Sets the lsb sublane at |pos| with least significant NBitsLeast of |in|
+    constexpr auto least(T in, u8 pos) const noexcept {
```
thecppzoo, line 231: Please don't use `u8` here.
If you look at the code that deals with indices, they are `int`.
The reason is that `int` is the natural integer size for the architecture, hence it minimizes the chances of penalties of integer size conversions.
Also, it is actually important, however subtle, to keep these *signed* integers, the reason being that integer overflow is "undefined behavior", this means the compiler is allowed to assume the program won't enter undefined behavior and thus won't overflow, then it can enable optimizations it otherwise would not be able to.
I tried to make an illustration of this but in 15 minutes I could not

inc/zoo/swar/SWAR.h
Scottbruceheart: fixed, pushed

inc/zoo/swar/SWAR.h
Scottbruceheart: fixed, pushed

Scottbruceheart:
rebased to trunk

inc/zoo/swar/SWAR.h
```diff
@@ -40,15 +40,19 @@ template<typename T> constexpr typename std::make_unsigned<T>::type lsbIndex(T v
 /// SWAR operations are usually constant time, log(lane count) cost, or O(lane count) cost.
 /// Certain computational workloads can be materially sped up using SWAR techniques.
 template<int NBits, typename T = uint64_t> struct SWAR {
+    static constexpr inline auto LaneCount = sizeof(T) * 8 / NBits;
```
thecppzoo, line 43: `Lanes` instead of `LaneCount`?

inc/zoo/swar/SWAR.h
```diff
@@ -74,57 +78,152 @@ template<int NBits, typename T = uint64_t> struct SWAR {
...
+
+    // Returns only the most significant bits at specified position.
+    constexpr auto most(u32 pos) const noexcept {
+        constexpr auto filter = (T(1) << LaneBits) - 1;
+        const auto keep = (filter << (LaneBits * pos)) & MostMask;
+        return this->m_v & keep;
+    }
+
+    // Returns only the most significant bits at specified position, 'decoded' to their integer value.
+    constexpr auto mostFlat(u32 pos) const noexcept {
```
thecppzoo, line 143: there is a missing operation here, `flattenMost`, which blits the most significant bits into the leasts

inc/zoo/swar/SWAR.h
```diff
@@ -74,57 +78,152 @@ template<int NBits, typename T = uint64_t> struct SWAR {
...
-  return (T(1)<<NBits)-1;
+constexpr auto leastNBitsMask() {
+    return (T(1)<<NBits)-1;
+  //return ~((~T(0))<<NBits);
+}
+
+template<int NBits, uint64_t T>
+constexpr auto leastNBitsMask() {
+  // return (u64(1)<<NBits)-1;
+    return ~((0ull)<<NBits);
```
thecppzoo, line 202: I prefer these to be constants instead of functions

inc/zoo/swar/SWAR.h
Scottbruceheart: Agree, useful op, but also needs HUGE WARNING about there being zero guarantee the sublanes will be the same size.

inc/zoo/swar/SWAR.h
Scottbruceheart: fixed

inc/zoo/swar/SWAR.h
Scottbruceheart: How do we make them constants for arbitrary types?

PR #18, opened 2022-08-10
[SWAR][RobinHood] Add strawman implementation of Robin Hood table.

Scottbruceheart:
Inserts/exists. Creates a valid psl/hash arrangement in its naive metadata.

Follow PR for translate to a version of simple swar metadata hash.

Adds mostSignificantBits function to SWAR.

### DIFF
--- a/inc/zoo/map/robinhood.h
+++ b/inc/zoo/map/robinhood.h
@@ -1,0 +1,145 @@
+#pragma once
+
+#include "zoo/swar/SWAR.h"
+
+#include <vector>
+
+namespace zoo {
+
+namespace rh {
+
+using u64 = uint64_t;
+using u32 = uint32_t;
+using u16 = uint16_t;
+using u8 = uint8_t;
+
+template<int NBits>
+constexpr auto cheapOkHash(u64 n) {
+    constexpr auto shift = (NBits * ((64 / NBits)-1));
+    constexpr u64 allOnes = meta::BitmaskMaker<u64, 1, NBits>::value;
+    auto temporary = allOnes * n;
+    auto higestNBits = temporary >> shift;
+    return (0 == (64 % NBits)) ? higestNBits : swar::isolate<NBits, u64>(higestNBits);
+}
+
+// The naive method of keeping metadata.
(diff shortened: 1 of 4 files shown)


inc/zoo/map/robinhood.h
```diff
@@ -0,0 +1,143 @@
...
+constexpr auto cheapOkHash(u64 n) {
+    constexpr auto shift = (NBits * ((64 / NBits)-1));
+    constexpr u64 allOnes = meta::BitmaskMaker<u64, 1, NBits>::value;
+    auto temporary = allOnes * n;
+    auto higestNBits = temporary >> shift;
+    return (0 == (64 % NBits)) ? higestNBits : swar::isolate<NBits, u64>(higestNBits);
+}
+
+// The naive method of keeping metadata.
+template<u8 PSLBits, u8 HashBits> struct StrawMetadata {
```
thecppzoo, line 26: Let's not make this a template, it's more useful as a concrete class

inc/zoo/map/robinhood.h
```diff
@@ -0,0 +1,143 @@
...
+      }
+    }
+
+    template <typename KeyCheck> 
+    std::pair<int, bool> insert(Key k, KeyCheck kc) {
+        u32 hashbits = reducedhash(k);
+        u32 slot = slotFromKey<u32>(k);
+        auto slotFind = findSlot(k, kc);
+        if (slotFind.second) {
+            return std::make_pair(slotFind.first, false);  // already in
```
thecppzoo, line 104: `return {slotFind.first, false};` <- since C++17 we have "CTAD", "Class Template Argument Deduction", many of these `make_...` were made obsolete

inc/zoo/map/robinhood.h
Scottbruceheart: Sure. Though I wrote this one wrong, I meant to ensure the functions returning psl/hash returned the right width.

inc/zoo/map/robinhood.h
Scottbruceheart: Old habits die eventually. :)

PR #22, opened 2022-08-14
WIP Basic matchAttempt vs a simple RH swar block.  

Scottbruceheart:
This is an ok impl.
I'd prefer to use the sublanes abstraction a bit more cleanly, and I feel like we're missing a trick with equality testing.
Added greaterEqualZeroLeft (since we call that in a few places, though opt compiler probably catches that, note we still don't have a general greaterEqual)
Added some basic tests for matchAttempt.


inc/zoo/map/RobinHood_util.h
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "zoo/meta/BitmaskMaker.h"
+#include "zoo/meta/log.h"
```
thecppzoo, line 4: These come from `SWAR.h`

inc/zoo/swar/SWAR.h
```diff
@@ -256,16 +256,29 @@ struct BooleanSWAR: SWAR<NBits, T> {
...
+    template<int NB, typename TT> friend
+    constexpr BooleanSWAR<NB, TT> differents(SWAR<NB, TT> left, SWAR<NB, TT> right) noexcept;
 
 };
 
+// Differents becuase "differences of lanes in boolean bits"
+template<int NB, typename TT> constexpr
+BooleanSWAR<NB, TT> differents(SWAR<NB, TT> left, SWAR<NB, TT> right) noexcept {
+    return left ^ right;
+}
```
thecppzoo, line 275: This does not leave the most significant bit set on "true", the MSB must be the boolean, and the rest 0

inc/zoo/map/RobinHood_util.h
Scottbruceheart: fixed.

Scottbruceheart:
This works for basic match, but your provider abstraction is better.

Scottbruceheart:
This is too old, closing.

PR #24, opened 2022-08-24
[SWAR] Compares for SWAR booleans.

Scottbruceheart:
Add some compares for swar boolean, including a 'left argument zero' swar compare.


Scottbruceheart:
This is too old, closing.

PR #25, opened 2022-08-24
[robinhood][swar]

Scottbruceheart:
Based off https://github.com/thecppzoo/zoo/pull/24

Contains matchattempt and basic organization of slots and basic separation of concerns around their matching.


inc/zoo/map/RobinHood_straw.h
```diff
@@ -0,0 +1,108 @@
+#pragma once
```
thecppzoo, line 1: Please let us not use `#pragma once`, the reason is that with proper header guards you can control from test drivers how the headers should look like, I mean overriding the headers actually used from within the test driver

inc/zoo/map/RobinHood_util.h
```diff
@@ -0,0 +1,145 @@
...
+    } else if constexpr(sizeof(Key) == 4) {
+        u32 r = k * FibonacciGoldenRatioReciprocal32; 
+        return r;
+    } else if constexpr(sizeof(Key) == 2) {
+        u16 r = k * FibonacciGoldenRatioReciprocal16;
+        return r;
+    } else if constexpr(sizeof(Key) == 1) {
+        u8 r = k * FibonacciGoldenRatioReciprocal8;
+        return r;
+    }
```
thecppzoo, line 56: Put these numbers in an array and index the array from `sizeof(`

inc/zoo/map/RobinHood_util.h
```diff
@@ -0,0 +1,145 @@
...
+             SM haystack, SM needleHashes, SM needlePSL) {
+        const auto haystackPSL = SM{SSL::LeastMask & haystack.value()};
+        const auto d = deadline(haystackPSL, needlePSL);  // breaks abstraction
+        const auto needle = needleHashes | needlePSL;
+
+        const auto matches = equals(haystack, needle);
+        const auto searchEnds = d ? 1 : 0; 
+        // Returned value is MSB boolean array with 'finality' bit on at
+        // position 0. Breaks if PSLs are width 1 (but so does everything else)
+        return SM{searchEnds | ((d - 1) & matches.value())};
```
thecppzoo, line 114: Valiant effort, I'm not patronizing.  It took me months of unconscious thinking to determine that a match attempt needs to return all potential matches and the deadline.
The use of the encoding will not be profitable compared to `std::tuple` of the two values described above.

inc/zoo/map/RobinHood_util.h
```diff
@@ -0,0 +1,145 @@
...
+        // holds.
+        auto satisfied = greaterEqual_MSB_off(pslHaystack, pslNeedle);
+        auto broken = not satisfied;
+        return swar::isolateLSB(broken.value());
+    }
+
+    // Has no intrinsic binding to the metadata, just easier to write with
+    // using decls.
+    static constexpr auto attemptMatch(
+             SM haystack, SM needleHashes, SM needlePSL) {
```
thecppzoo, line 105:
```suggestion
    static constexpr auto attemptMatch(
        SM haystack, SM needleHashes, SM needlePSL
     ) {
```
I believe "accidental" indentation levels is counterproductive

inc/zoo/map/RobinHood_util.h
```diff
@@ -0,0 +1,145 @@
...
+}
+
+template<typename Key>auto slotFromKeyUnitary(Key k) noexcept {
+    return 1;
+}
+
+template<int NBitsHash, int NBitsPSL, typename T = u64> struct SlotOperations {
+    using SSL = swar::SWARWithSubLanes<NBitsHash, NBitsPSL, T>;
+    using SM = swar::SWAR<NBitsHash+ NBitsPSL, T>;
+    using BoolSM = swar::BooleanSWAR<NBitsHash+ NBitsPSL, T>;
```
thecppzoo, line 78: I object to this name.
It seems this class is used as a namespace (not the syntactical feature but the concept).

inc/zoo/map/RobinHood_util.h
```diff
@@ -0,0 +1,145 @@
...
+        // Returned value is MSB boolean array with 'finality' bit on at
+        // position 0. Breaks if PSLs are width 1 (but so does everything else)
+        return SM{searchEnds | ((d - 1) & matches.value())};
+    }
+};
+
+/// Slot metadata provides the affordances of 
+///     'attempt to match this needle of hashes and PSLs'
+/// Contains sizeof(T)/(NBitsHash,NBitsPSL) hashes and PSLs
+template<int NBitsHash, int NBitsPSL, typename T = u64> struct SlotMetadata {
```
thecppzoo, line 121: In a parallel implementation I use `Metadata` *inside* a Robin-Hood specific class.
`SlotMetadata` seems a class subordinate to a class or concept of entry/slot; but this does not work, because the metadata is a collection of entries.

Scottbruceheart:
This is too old and all functionality has since been implemented, closing.

PR #29, opened 2022-08-25
Em/rh match

thecppzoo:
As PR


inc/zoo/map/RobinHood.h
```diff
@@ -0,0 +1,412 @@
...
+                if(kc(matchIndex)) {
+                    return std::tuple(true, matchIndex);
+                }
+                positives = Metadata{clearLSB(positives.value())};
+            }
+            if(deadline) {
+                auto position = index + Metadata{deadline}.lsbIndex();
+                return std::tuple(false, position);
+            }
+            // Skarupke's tail allows us to not have to worry about the end
```
Scottbruceheart, line 225: I do think we need to handle tail and no tail at some point in the future (since no tail might be better for smaller tables). That's after presentation though, I suspect.

Scottbruceheart: And shouldn't the metdata provider have the concept of the skarupke tail, not RH_Backend? Would that require it keeping track of what index results are in (instead of RH_Backend)

inc/zoo/map/RobinHood.h
```diff
@@ -0,0 +1,412 @@
...
+            // of the metadata
+            ++p;
+            index += Metadata::NSlots;
+            needle = needle + Progression;
+        }
+    }
+        
+    template<typename KeyComparer>
+    constexpr auto
+    findThroughIndirectJump(
```
Scottbruceheart, line 235: why does misalignment only go to 7?

inc/zoo/map/RobinHood.h
```diff
@@ -0,0 +1,412 @@
...
+            higestNBits :
+            swar::isolate<NBits>(higestNBits);
+}
+
+
+template<typename T>
+constexpr auto fibonacciIndexModulo(T index) {
+    constexpr std::array<uint64_t, 4>
+        GoldenRatioReciprocals = {
+            159,
```
Scottbruceheart, line 310: add comment for size

inc/zoo/map/RobinHood.h
```diff
@@ -0,0 +1,412 @@
...
+            WithTail +
+            MD::NSlots - 1 // to calculate the ceiling rounding
+        ) / MD::NSlots
+    ;
+    constexpr static inline auto SlotCount = SWARCount * MD::NSlots;
+
+    std::array<MD, SWARCount> md_;
+    using value_type = std::pair<K, MV>;
+
+    /// \todo Scatter key and value in a flavor
```
Scottbruceheart, line 357: what is a flavor

inc/zoo/swar/SWAR.h
```diff
@@ -346,26 +361,26 @@ greaterEqual_MSB_off(SWAR<NBits, T> left, SWAR<NBits, T> right) noexcept {
...
 constexpr auto
 equals(SWAR<NBits, T> a1, SWAR<NBits, T> a2) {
-    not differents(a1, a2);
+    return not differents(a1, a2);
 }
 
 /*
 This is just a draft implementation:
-1. The isolator needs pre-computing instead of adding 3 ops per iteration
+b1. The isolator needs pre-computing instead of adding 3 ops per iteration
```
Scottbruceheart, line 383: looks like a typo?

Scottbruceheart (approved):
We need to add tests, but looks strong enough.

inc/zoo/map/RobinHood.h
thecppzoo: The metadata provider ought not to have knowledge of the tail: it ought to only have knowledge of the size of the metadata it can provide.
The type of tail is directly bound to the meaning of the flavor of the RH table, then I think the natural place is the backend

inc/zoo/map/RobinHood.h
thecppzoo: 0 to 7 assuming a max count of 8 elements (hard-coded for testing purposes)

PR #32, opened 2022-09-10
SWAR insertion

thecppzoo:
First cut of insertion into Robin Hood table via swar methods.
First figures out the chain of insertions that need to happen (to some fixed maximum), then applies those insertions.
Fairly involved.

### DIFF
--- a/inc/zoo/map/RobinHood.h
+++ b/inc/zoo/map/RobinHood.h
@@ -39,7 +39,8 @@ struct MisalignedGenerator {
             // I'd prefer to not use std::make_unsigned_t, since how do we
             // "make unsigned" user types?
         auto firstPartLowered = firstPart.value() >> MisalignmentBits;
-        auto secondPartRaised = secondPart.value() << (Width - MisalignmentBits);
+        auto secondPartRaised =
+            secondPart.value() << (Width - MisalignmentBits);
         return T{firstPartLowered | secondPartRaised};
     }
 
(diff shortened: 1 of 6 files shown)


inc/zoo/map/RobinHood.h
```diff
@@ -262,32 +263,42 @@ struct RH_Backend {
...
         }
     }
+
+    template<typename Inserter>
+    constexpr auto
+    evictionChain(Inserter, int index) {
+        auto baseIndex = index / Metadata::Lanes;
+        constexpr auto Ones = meta::BitmaskMaker<U, 1, Width>::value;
+        constexpr auto Progression = Metadata{Ones * Ones};
+    }
```
thecppzoo, line 301: Junk, please ignore
```suggestion
```

inc/zoo/map/RobinHood.h
```diff
@@ -385,23 +422,182 @@ struct RH_Frontend_WithSkarupkeTail {
...
-        auto keyChecker =
-            [thy = this, &k](size_t ndx) noexcept {
-                return KE{}(thy->values_[ndx].template as<value_type>()->first, k);
-            };
+    auto findParameters(const K &k) const noexcept {
         auto hashCode = Hash{}(k);
         auto fibonacciScrambled = fibonacciIndexModulo(hashCode);
         auto homeIndex =
             lemireModuloReductionAlternative<RequestedSize>(fibonacciScrambled);
         auto hoisted = hashReducer<HashBits>(hashCode);
```
thecppzoo, line 430: This is not what we want, but to use either fibonacciScrambled or something

PR #34, opened 2022-09-11
Insertion improvements (tests)

thecppzoo:
Material reduction in instructions in inner find loop by removing the check against deadlines: the PSL growth patterns mean that any deadline breach will result in no full match from calling equals(haystack, needle).
Fixes a few basic tests related to matching.
Switches SWARWithSublanes to <least, most> standard on specifying bit positions.

### DIFF
--- a/inc/zoo/map/RobinHood.h
+++ b/inc/zoo/map/RobinHood.h
@@ -16,7 +16,6 @@
     #include <assert>
 #endif
 
-template<typename> struct GT;
 
 namespace zoo {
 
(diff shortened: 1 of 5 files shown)


inc/zoo/map/RobinHood.h
```diff
@@ -16,6 +16,8 @@
     #include <assert>
 #endif
 
+template<typename> struct GT;
```
thecppzoo, line 19:
```suggestion
```

PR #41, opened 2022-09-13
Implements sublanes as a composition of swar

thecppzoo:


Scottbruceheart:
I think sublanes can be 1 bit wide (and not fail), so I don't know if composing swars to implement them will be effective.
Stated differently: the seamntics of a sublane might be different than that of a lane.

PR #42, opened 2022-09-13
Guards against exhaustion of stack for relocations or PSL

thecppzoo:
Insertion with journaled metadata and value writing. Tests for same.

### DIFF
--- a/inc/zoo/map/RobinHood.h
+++ b/inc/zoo/map/RobinHood.h
@@ -130,7 +130,8 @@ struct RH_Backend {
         auto needle = makeNeedle(0, hoistedHash);
 
         for(;;) {
-            auto result = potentialMatches(needle, *p);
+            auto hay = *p;
+            auto result = potentialMatches(needle, hay);
             auto positives = result.potentialMatches;
             while(positives.value()) {
                 auto matchSubIndex = positives.lsbIndex();
@@ -366,12 +367,16 @@ struct RH_Frontend_WithSkarupkeTail {
                 }
                 // the last element is special because it is a
                 // move-construction, not a move-assignment
-                auto fromIndex = relocations[--relocationsCount];
+                --relocationsCount;
+                auto fromIndex = relocations[relocationsCount];
                 values_[index].build(
                     std::move(values_[fromIndex].value())
                 );
-                auto mde = newElements[relocationsCount];
-                md_[index] = md_[index].blitElement(intraIndex, mde);
+                md_[swarIndex] =
+                    md_[swarIndex].blitElement(
+                        intraIndex, elementToInsert
+                    );
+                elementToInsert = newElements[relocationsCount];
                 index = fromIndex;
                 swarIndex = index / MD::NSlots;
                 intraIndex = index % MD::NSlots;
@@ -380,15 +385,15 @@ struct RH_Frontend_WithSkarupkeTail {
                     fromIndex = relocations[relocationsCount];
                     values_[index].value() =
                         std::move(values_[fromIndex].value());
-                    mde = newElements[relocationsCount];
-                    md_[index] = md_[index].blitElement(intraIndex, mde);
+                    md_[swarIndex] = md_[swarIndex].blitElement(intraIndex, elementToInsert);
+                    elementToInsert = newElements[relocationsCount];
                     index = fromIndex;
                     swarIndex = index / MD::NSlots;
                     intraIndex = index % MD::NSlots;
                 }
                 values_[index].value() = std::pair(k, mv);
                 md_[swarIndex] =
-                    md_[swarIndex].blitElement(intraIndex, mde);
+                    md_[swarIndex].blitElement(intraIndex, elementToInsert);
                 return std::pair{values_.data() + index, true};
             }
             if(HighestSafePSL < evictedPSL) {
@@ -404,6 +409,7 @@ struct RH_Frontend_WithSkarupkeTail {
 
             // "push" the index of the element that will be evicted
             relocations[relocationsCount] = index;
+            // we have a place for the element being inserted, at this index
             newElements[relocationsCount++] = elementToInsert;
             if(MaxRelocations <= relocationsCount) {
                 throw RelocationStackExhausted("");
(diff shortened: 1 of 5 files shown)


inc/zoo/map/RobinHood.h
```diff
@@ -18,9 +18,19 @@
...
 namespace rh {
 
+struct RobinHoodException: std::runtime_error {
+    using std::runtime_error::runtime_error;
+};
+
+struct MaximumProbeSequenceLengthExceeded: RobinHoodException {
+    using RobinHoodException::RobinHoodException;
+};
+struct RelocationStackExhausted: RobinHoodException {
```
Scottbruceheart, line 30: Nit: More table size exhausted than a stack exahusted?

inc/zoo/map/RobinHood.h
```diff
@@ -386,26 +337,34 @@ struct RH_Frontend_WithSkarupkeTail {
         auto swarIndex = index / MD::Lanes;
         auto intraIndex = index % MD::Lanes;
         auto mdp = this->md_.data() + swarIndex;
-        constexpr auto MaxRelocations = 64;
+        constexpr auto MaxRelocations = 100;
```
Scottbruceheart, line 340: If we relocate too may times...we throw an exception? Fall back to something else?

inc/zoo/map/RobinHood.h
thecppzoo: Table size exhausted is actually not an exception, since the table is such that first the encoding for the PSL is exhausted.
I have to remove the sentinel part in the code.

inc/zoo/map/RobinHood.h
thecppzoo: We have to think deeper about whether or how to support the strong exception guarantee.
It is expensive to determine if the insertion will succeed, it is expensive to have a journal of operations to do to commit an insertion.
This is good to expose to the audience of the presentations.
Now, we have a solution, that is a compromise between the two options.

inc/zoo/map/RobinHood.h
Scottbruceheart: If we insert with a degenerate hash function that always hits the last element of the table, then we'll exhaust psl as we hit the last element of the table, so I suppose that is correct.

inc/zoo/map/RobinHood.h
Scottbruceheart: I thought we had inplace insert implemented (though not tested well) and now journaled insert implemeted and tested.  Seeems like we should keep both functions around, and (if time allows) fix up insert and benchmark and present results for both.

inc/zoo/map/RobinHood.h
```diff
@@ -102,71 +112,6 @@ struct RH_Backend {
         return startingPSLmadePotentialPSLs;
     }
 
-    template<int Misalignment, typename KeyComparer>
```
Scottbruceheart, line 105: If unaligned find is easy (which it is), we should keep both styles of find around, present both, and present benchmarks on both (if time allows)

inc/zoo/map/RobinHood.h
```diff
@@ -387,62 +336,89 @@ struct RH_Frontend_WithSkarupkeTail {
...
-            auto evictedPSL = mdBackup.PSLs().at(intraIndex);
+            auto md = *mdp;
+            auto evictedPSL = md.PSLs().at(intraIndex);
             if(0 == evictedPSL) { // end of eviction chain!
-                assignMetadataElement(
-                    deadline,
-                    elementToInsert.shiftLanesLeft(intraIndex),
-                    mdp
-                );
+                if(SlotCount - 1 <= index) {
```
Scottbruceheart, line 357: Name unclear: is index here a global index, a swar index, or a local offset?

inc/zoo/map/RobinHood.h
```diff
@@ -387,62 +336,89 @@ struct RH_Frontend_WithSkarupkeTail {
...
+                    throw MaximumProbeSequenceLengthExceeded("full table");
+                }
                 if(0 == relocationsCount) { // direct build of a new value
-                    values_[index].build(std::pair{k, mv}); // inplace?
+                    values_[index].build(
+                        std::piecewise_construct,
```
Scottbruceheart, line 362: nit: at some points exact semantics of why build vs why not should be in a comment somewhere.

inc/zoo/map/RobinHood.h
```diff
@@ -387,62 +336,89 @@ struct RH_Frontend_WithSkarupkeTail {
...
                     index = fromIndex;
+                    swarIndex = index / MD::NSlots;
+                    intraIndex = index % MD::NSlots;
                 }
                 values_[index].value() = std::pair(k, mv);
+                md_[swarIndex] =
+                    md_[swarIndex].blitElement(intraIndex, elementToInsert);
                 return std::pair{values_.data() + index, true};
             }
+            if(HighestSafePSL < evictedPSL) {
```
Scottbruceheart, line 400: HighestSafePSL = (1 << PSL_Bits) - MD::NSlots - 1;
5 bits 8 slots is 32-9
4 bits 8 slots is 16-9
3 bits 8 slots is 8-9
This seems ok for now, but we'll have to do something explicit to support small PSL in future, probably.

Scottbruceheart: Alternately stated: I think this constrains our max load factor heavily.

inc/zoo/map/RobinHood.h
```diff
@@ -387,62 +336,89 @@ struct RH_Frontend_WithSkarupkeTail {
...
-            );
-            // now the insertion will be for the old metadata entry
-            elementToInsert = MD(mdBackup.hashes().at(intraIndex));
             // "push" the index of the element that will be evicted
-            relocations[relocationsCount++] = index;
+            relocations[relocationsCount] = index;
+            // we have a place for the element being inserted, at this index
+            newElements[relocationsCount++] = elementToInsert;
+            if(MaxRelocations <= relocationsCount) {
+                throw RelocationStackExhausted("");
```
Scottbruceheart, line 416: We need a solution at some point in the future for relocation stack exhaustion (of some sort).  It'll constrain maximum load factor for tables, esp large ones.

inc/zoo/map/RobinHood.h
thecppzoo: Sure, I like the 'optimistic' insertion implementation.

inc/zoo/map/RobinHood.h
thecppzoo: This code drifted so much from the head that it needed removal to prevent confusion

inc/zoo/map/RobinHood.h
thecppzoo: within the code, I attempt to use these consistently:
`index`: the global index of elements
`swarIndex`,
`intraIndex`

inc/zoo/map/RobinHood.h
thecppzoo: It is actually commented: "direct build of a new value", in the other place where a construction happens, it is also explained why,

inc/zoo/map/RobinHood.h
```diff
@@ -387,62 +336,89 @@ struct RH_Frontend_WithSkarupkeTail {
...
                     return std::pair{values_.data() + index, true};
                 }
                 // the last element is special because it is a
                 // move-construction, not a move-assignment
-                auto fromIndex = relocations[--relocationsCount];
+                --relocationsCount;
+                auto fromIndex = relocations[relocationsCount];
                 values_[index].build(
                     std::move(values_[fromIndex].value())
                 );
```
thecppzoo, line 375: Here is the other place: move construction (build)

inc/zoo/map/RobinHood.h
thecppzoo: There is something here: for tables so small that the number of lanes in a SWAR has any significance, the whole of a SWAR implementation is counterproductive.
I don't see the argument you make:
For a number of lanes of 8 per SWAR, 8 codes from 2^`PSL_Bits` are unsafe, codes in which you simply can't use SWAR, if we have to test for those, I think the performance would crater

inc/zoo/map/RobinHood.h
thecppzoo: For demonstration purposes we can use however large a factor let's us do the work.
This is something that depends in an unresolved question, of journal/no journal; strong/not strong exception guarantees.

PR #43, opened 2022-09-14
[robin hood][swar] Testing for robin hood find. Expose dependence on undefined shift behavior, fix.

Scottbruceheart:
Undefined behavior of type << <typewidth> is indeed very bad. Avoid that by shifting n-1, then 1.  Fix usage in two places: misalignment generator and realignment of deadline/metadata post-find.

Add a few related functions.

Add some tests cases around finding vs artificially constructed (but valid) metadata.

### DIFF
--- a/inc/zoo/map/RobinHood.h
+++ b/inc/zoo/map/RobinHood.h
@@ -16,7 +16,6 @@
     #include <assert>
 #endif
 
-
 namespace zoo {
 
 namespace rh {
@@ -164,6 +163,10 @@ struct RH_Backend {
         __builtin_unreachable();
     }
 
+    // Returned metadata is all garbage except for any entry in a deadline,
+    // which contains the PSL and hash of the element which matches the
+    // deadline. IE: it is returnedMetadata.at(deadline.lsbindex()) that is the
+    // PSL+hashbits
     template<typename KeyComparer>
     constexpr auto
     findMisaligned_assumesSkarupkeTail(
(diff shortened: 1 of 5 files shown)


inc/zoo/map/RobinHood.h
```diff
@@ -164,6 +163,10 @@ struct RH_Backend {
         __builtin_unreachable();
     }
 
+    // Returned metadata is all garbage except for any entry in a deadline,
```
thecppzoo, line 166:
```suggestion
    // Returned metadata is irrelevant except for the element at the same index as the deadline
```

inc/zoo/swar/SWAR.h
```diff
@@ -71,6 +71,13 @@ struct SWAR {
     SWAR_BINARY_OPERATORS_X_LIST
     #undef X
 
+    // Returns lane at position with other lanes cleared.
+    constexpr T lane(int position) const noexcept {
```
thecppzoo, line 75:
```suggestion
    constexpr T isolateLane(int position) const noexcept {
```

inc/zoo/swar/SWAR.h
```diff
@@ -105,6 +112,15 @@ struct SWAR {
...
     }
 
+    // This will shift 1 bit less right to avoid undefined behavior. Use only
+    // in conditions when the caller can prove that laneCount is not zero.
+    constexpr SWAR shiftLanesRightSafe(int laneCount) const noexcept {
+        return SWAR(value() >> ((NBits * laneCount) - 1));
+    }
+    constexpr SWAR shiftOneBitRight() const noexcept {
+        return SWAR(value() >> 1);
+    }
```
thecppzoo, line 122: please eliminate

inc/zoo/map/RobinHood.h
```diff
@@ -207,18 +210,16 @@ struct RH_Backend {
...
                 auto mdd = Metadata{deadline};
                 auto toAbsolute = [](auto v, auto ma) {
-                    return
-                        Metadata{
-                            v.shiftLanesLeft(ma) |
-                            v.shiftLanesRight(Metadata::NSlots - ma)
-                        };
+                    auto shiftedLeft = v.shiftLanesLeft(ma);
+                    auto shiftedRight =
+                        v.shiftLanesRightSafe(Metadata::NSlots - ma).shiftOneBitRight();
```
thecppzoo, line 221:
```suggestion
                        v.shiftLanesRight(Metadata::NSlots - ma - 1).shiftLanesRight(1);
```

test/map/RobinHood.test.cpp
```diff
@@ -70,7 +60,7 @@ std::ostream &operator<<(std::ostream &out, V v) {
             if(v.intraIndex == ndx) {
                 ptr += sprintf(ptr, "<");
             }
-            ptr += sprintf(ptr, "%02llx", val & 0xFF);
+            ptr += sprintf(ptr, "%02llx", val & 0xFFull);
```
thecppzoo, line 63: please don't be putting `ull` where is not needed

thecppzoo (changes requested):
We also ahve to fix `toAbsolute`

PR #46, opened 2022-09-15
First benchmark

thecppzoo:
Initial benchmarks between rh, unordered map and map.  Large number of strings from a real text, count occurrence.


inc/zoo/map/RobinHood.h
```diff
@@ -222,7 +222,7 @@ template<
     typename KE = std::equal_to<K>,
     typename U = std::uint64_t,
     typename Scatter = FibonacciScatter<U>,
-    typename RangeReduce = LemireReduce<RequestedSize, U>,
+    typename RangeReduce = LemireReduce<RequestedSize_, U>,
```
Scottbruceheart, line 225: why?

PR #64, opened 2023-12-24
Saturated Addition, Unsigned.

Scottbruceheart:


inc/zoo/swar/associative_iteration.h
```diff
@@ -106,6 +106,36 @@ fullAddition(SWAR<NB, B> s1, SWAR<NB, B> s2) {
     return { result, BS{carry.value()}, BS{overflow.value()} };
 };
 
+static_assert( SWAR<4, u32>(0x1111'1111).value() == fullAddition(SWAR<4, u32>(0x0111'1101), SWAR<4, u32>(0x1000'0010)).result.value());
```
thecppzoo, line 109: At this point of maturity of the library, I don't see the point in continuing to say `SWAR<4, uint32_t>` in the test files.  I think we can simply get away with defining however `using S4_32 = SWAR<4, uint32_t>`  we want type aliases globally in the test file; I advocate for the convention of `S` followed by the number of bits per lane, `_` and the number of bits in the unsigned integer we mean.
And we should remove `static_assert` from the user-facing headers.

inc/zoo/swar/associative_iteration.h
```diff
@@ -106,6 +106,36 @@ fullAddition(SWAR<NB, B> s1, SWAR<NB, B> s2) {
     return { result, BS{carry.value()}, BS{overflow.value()} };
 };
 
+static_assert( SWAR<4, u32>(0x1111'1111).value() == fullAddition(SWAR<4, u32>(0x0111'1101), SWAR<4, u32>(0x1000'0010)).result.value());
+static_assert( SWAR<4, u32>(0x0000'0000).value() == fullAddition(SWAR<4, u32>(0x0111'1101), SWAR<4, u32>(0x1000'0010)).carry.value());
+static_assert( SWAR<4, u32>(0x0000'0000).value() == fullAddition(SWAR<4, u32>(0x0111'1101), SWAR<4, u32>(0x1000'0010)).overflow.value());
+
+template<int NB, typename B>
+constexpr ArithmeticResultTriplet<NB, B>
+fullSaturatedAddition(SWAR<NB, B> s1, SWAR<NB, B> s2) {
```
thecppzoo, line 115: I think there are two flavors of saturation: unsigned and signed.

thecppzoo: What about clamping?

inc/zoo/swar/associative_iteration.h
Scottbruceheart: SX_Y meaning bits per lane and bits in fundamental type is a good convention.

inc/zoo/swar/associative_iteration.h
Scottbruceheart: Name fixed, clamping is for a future PR.

Scottbruceheart:
too diverged from main, making clean PR

PR #68, opened 2024-02-01
Em/pokerbotic subsumption

thecppzoo:


jamierpond:
curious why to merge pokerbotic into this repo? does it not make sense as a standalone repo? one thought is that it could make sense to keep zoo more focused, but i don't have all the context that you do @thecppzoo

thecppzoo:
> curious why to merge pokerbotic into this repo? does it not make sense as a standalone repo? one thought is that it could make sense to keep zoo more focused, but i don't have all the context that you do @thecppzoo

Discussed out of band: Pokerbotic is a good demo of the SWAR techniques, and a good application to drive further development

PR #69, opened 2024-02-06
Demos of the SWAR library

thecppzoo:
Reviewers: please, in the case you're satisfied with the PR up to a certain commit, please state your satisfaction in a comment.  If the other reviewer is also satisfied as of the same commit as you, then please merge up to that commit to the master branch.  This is more "agile"

### DIFF
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -23,7 +23,9 @@ add_subdirectory(dependencies/google_benchmark)
 
 add_executable(
   catch2Benchmark
-    catch2BenchmarkMain.cpp catch2Functions.cpp egyptian.cpp
+    catch2BenchmarkMain.cpp catch2Functions.cpp catch2swar-demo.cpp
+    atoi.cpp
+    egyptian.cpp
     RobinHood.benchmark.cpp
 )
 
@@ -32,7 +34,7 @@ add_executable(
 )
 
 add_executable(
-    zoo-atoi-benchmark benchmark_main.cpp atoi.cpp
+    zoo-atoi-benchmark benchmark_main.cpp bm-swar.cpp atoi.cpp
 )
 
 target_link_libraries(zoo-google-benchmark benchmark::benchmark)
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -1,8 +1,9 @@
-cmake_minimum_required (VERSION 2.6)
+cmake_minimum_required (VERSION 3.8)
 
-set(CMAKE_CXX_FLAGS "-std=c++17 -g3 -fno-unroll-loops")
+set(CMAKE_CXX_FLAGS "-fno-unroll-loops")
 
-set(CMAKE_BUILD_TYPE "Release")
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_BUILD_TYPE "RelWithDebInfo")
 
 project(ZooBenchmark)
 
@@ -16,13 +17,17 @@ include_directories(
 
 # build google benchmark (target: benchmark)
 # do not build tests of benchmarking lib
-# set(BENCHMARK_ENABLE_TESTING OFF CACHE BOOL "Suppressing benchmark's tests" FORCE)
-
-# add_subdirectory(dependencies/google_benchmark)
-
+set(BENCHMARK_ENABLE_TESTING OFF CACHE BOOL "Suppressing benchmark's tests" FORCE)
+add_subdirectory(dependencies/google_benchmark)
 
 add_executable(
   catch2Benchmark
     catch2BenchmarkMain.cpp catch2Functions.cpp egyptian.cpp
     RobinHood.benchmark.cpp
 )
+
+add_executable(
+    zoo-google-benchmark benchmark_main.cpp cfs.cpp cfs/cfs_utility.cpp
+)
+
+target_link_libraries(zoo-google-benchmark benchmark::benchmark)
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -15,6 +15,18 @@ if(CMAKE_BUILD_TYPE MATCHES "Release|RelWithDebInfo")
     endif()
 endif()
 
+# Macro to set properties for Xcode targets
+macro(set_xcode_properties TARGET_NAME)
+    if(CMAKE_GENERATOR STREQUAL Xcode)
+        set_target_properties(${TARGET_NAME} PROPERTIES
+            XCODE_ATTRIBUTE_ENABLE_AVX YES
+            XCODE_ATTRIBUTE_ENABLE_AVX2 YES
+            XCODE_ATTRIBUTE_OTHER_CPLUSPLUSFLAGS "-mavx -mavx2"
+            XCODE_ATTRIBUTE_OTHER_CFLAGS "-mavx -mavx2"
+        )
+    endif()
+endmacro()
+
 include_directories(
     ../test/inc
     dependencies/google_benchmark/include
@@ -36,14 +48,17 @@ add_executable(
     egyptian.cpp
     # RobinHood.benchmark.cpp
 )
+set_xcode_properties(catch2Benchmark)
 
 add_executable(
     zoo-google-benchmark benchmark_main.cpp cfs.cpp cfs/cfs_utility.cpp
 )
+set_xcode_properties(zoo-google-benchmark)
 
 add_executable(
     zoo-atoi-benchmark benchmark_main.cpp bm-swar.cpp atoi.cpp
 )
+set_xcode_properties(zoo-atoi-benchmark)
 
 target_link_libraries(zoo-google-benchmark benchmark::benchmark)
 target_link_libraries(zoo-atoi-benchmark benchmark::benchmark)
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -18,6 +18,7 @@ include_directories(
 # build google benchmark (target: benchmark)
 # do not build tests of benchmarking lib
 set(BENCHMARK_ENABLE_TESTING OFF CACHE BOOL "Suppressing benchmark's tests" FORCE)
+set(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING "Build type" FORCE)
 add_subdirectory(dependencies/google_benchmark)
 
 add_executable(
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -9,7 +9,7 @@ project(ZooBenchmark)
 
 include_directories(
     ../test/inc
-    ../../benchmark/include
+    dependencies/google_benchmark/include
     ../inc
     ../junkyard/inc
     ../test/third_party/Catch2/single_include
@@ -30,4 +30,9 @@ add_executable(
     zoo-google-benchmark benchmark_main.cpp cfs.cpp cfs/cfs_utility.cpp
 )
 
+add_executable(
+    zoo-atoi-benchmark benchmark_main.cpp atoi.cpp
+)
+
 target_link_libraries(zoo-google-benchmark benchmark::benchmark)
+target_link_libraries(zoo-atoi-benchmark benchmark::benchmark)
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -1,11 +1,19 @@
 cmake_minimum_required (VERSION 3.8)
 
+project(ZooBenchmark)
+
+
 set(CMAKE_CXX_FLAGS "-fno-unroll-loops")
 
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_BUILD_TYPE "RelWithDebInfo")
 
-project(ZooBenchmark)
+if(CMAKE_BUILD_TYPE MATCHES "Release|RelWithDebInfo")
+    if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
+        string(APPEND CMAKE_CXX_FLAGS_RELEASE " -march=native -fno-unroll-loops")
+        string(APPEND CMAKE_CXX_FLAGS_RELWITHDEBINFO " -march=native -fno-unroll-loops")
+    endif()
+endif()
 
 include_directories(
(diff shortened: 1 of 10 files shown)


benchmark/atoi-corpus.h
```diff
@@ -0,0 +1,70 @@
...
+            if(--count) { break; }
+            base += 9;
+        }
+        return Corpus8DecimalDigits(inputs, allCharacters);
+    }
+
+    struct Iterator {
+        Corpus8DecimalDigits *thy;
+        int *ip;
+        char *cp;
```
jamierpond, line 38: I don't understand the names of these variables. Are there more friendly names for these possible?

benchmark/bm-swar.cpp
```diff
@@ -0,0 +1,34 @@
...
+template<typename CorpusMaker, typename Callable>
+void runBenchmark(benchmark::State &s) {
+    std::random_device rd;
+    std::mt19937 g(rd());
+    auto corpus = CorpusMaker::makeCorpus(g);
+    Callable function;
+    for(auto _: s) {
+        goOverCorpus(corpus, function);
+    }
+}
```
jamierpond, line 27: nice

benchmark/catch2swar-demo.cpp
```diff
@@ -0,0 +1,37 @@
...
+    REQUIRE(fromLIBC == fromZoo);
+
+    auto haveTheRoleOfMemoryBarrier = -1;
+    #define X(Type, Fun) \
+        BENCHMARK(#Type) { \
+            return \
+                traverse(corpus, Invoke##Type{}, haveTheRoleOfMemoryBarrier); \
+        };
+    PARSE8BYTES_CORPUS_X_LIST
+    #undef X
```
jamierpond, line 35: Could you help me understand about these defines in the benchmarks?

vscode/zoo.code-workspace
```diff
@@ -0,0 +1,11 @@
+{
```
jamierpond, line 1: might be good to .gitignore vscode files if they are just ide config files?

benchmark/atoi-corpus.h
thecppzoo: Integer pointer and character pointer.
I have the policy of choosing deliberately short identifiers, especially if they feel cryptic, to indicate "this name does not matter".  These names do not matter, since their role is immediately apparent by each use.  Please accept this element of style and "resolve" this comment to indicate it.

benchmark/catch2swar-demo.cpp
thecppzoo: This is the best explanation in case you need to understand what an X macro is:
https://www.youtube.com/watch?v=hXZrGh9kCVs
I will document this separately, you may take the rain check and mark this as "resolved", or not.

vscode/zoo.code-workspace
thecppzoo: I am finding that vscode is such a poor tool that it makes sense to keep the configuration under version control.
vscode in general does not work if the configuration for a repository is outside of the repository, this should not be the case, but then it glitches.
Also, I'm using vscode due to its popularity.  I want to prove the repository is compatible.
I prefer this stays as it is.  Do you agree? mark "resolved" in this case.

benchmark/CMakeLists.txt
```diff
@@ -15,6 +15,18 @@ if(CMAKE_BUILD_TYPE MATCHES "Release|RelWithDebInfo")
     endif()
 endif()
 
+# Macro to set properties for Xcode targets
+macro(set_xcode_properties TARGET_NAME)
+    if(CMAKE_GENERATOR STREQUAL Xcode)
+        set_target_properties(${TARGET_NAME} PROPERTIES
+            XCODE_ATTRIBUTE_ENABLE_AVX YES
+            XCODE_ATTRIBUTE_ENABLE_AVX2 YES
+            XCODE_ATTRIBUTE_OTHER_CPLUSPLUSFLAGS "-mavx -mavx2"
```
jamierpond, line 24: will run this on my M1 machine later on today, but i think you'll need another layer here of 
```cmake
if (${TARGET_PLATFORM} == X86_64) 
    // enable avx
endif()
```
will find the code i've used before which will do this only when AVX is actually detected on the machine, but i think this will need to be updated to work on ARM

benchmark/atoi.cpp
```diff
@@ -1,6 +1,8 @@
 #include "zoo/swar/SWAR.h"
 #include "zoo/swar/associative_iteration.h"
 
+#include <immintrin.h>
```
jamierpond, line 4: might be good to ifdef here based on platform?

benchmark/CMakeLists.txt
thecppzoo: Wait, doesn't XCode in M1 allow you to cross compile transparently to x86-64?

benchmark/atoi.cpp
thecppzoo: This is just an experiment.

benchmark/CMakeLists.txt
jamierpond: I think it would be cool to try and keep the code (where possible) architecture independent, otherwise could become a pain to try and go back and wrangle it to work in the future.

benchmark/atoi-corpus.h
jamierpond: I tend to advocate for verbose names, even if that is something like `tempVal` (I'm sure you could think of something better than `tempVal`), otherwise readers might think they are missing some context. However, if this is already and established convention in the code, then it should probably continue. Just my two cents!

benchmark/CMakeLists.txt
Scottbruceheart: Keeping code as arch independent as possible seems critical if our objectives are to work everywhere. Which I think they are.

benchmark/atoi.cpp
Scottbruceheart: atoi entirely is an experiment?

benchmark/CMakeLists.txt
thecppzoo: Yes, friends, architecture independence is a primary objective.  But we have the freedom to draft and experiment architecture specific.

benchmark/atoi.cpp
thecppzoo: yes, there are several atoi experiments to compare and see what affects performance at a micro level.

benchmark/atoi-corpus.h
```diff
@@ -0,0 +1,135 @@
...
+    {}
+
+    template<typename G>
+    static auto makeCorpus(G &generator) {
+        auto count = 1031; // 1031 is a prime number, this helps to disable in
+        // practice the branch predictor, the idea is to measure the performance
+        // of the code under measurement, not how the the unrealistic conditions
+        // of microbenchmarking help/hurt the code under measurement
+        std::string allCharacters;
+        allCharacters.resize(count * 9);
```
Scottbruceheart, line 24: why 9 and not a prime?

benchmark/atoi-corpus.h
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    struct Iterator {
+        Corpus8DecimalDigits *thy;
+        int *ip;
+        char *cp;
+
+        Iterator &operator++() {
+            ++ip;
+            cp += 9;
```
Scottbruceheart, line 45: magic numbers are magic

benchmark/atoi-corpus.h
thecppzoo: "otherwise readers might think":
Unfortunately, coding with an audience in mind that may include very superficial readers is limitative, it is a drag on the authors and contributors.
This codebase requires readers to understand the concept of negative space, things that may even be conspicuous because of their absence.
There used to be a great presentation about Post Modern C++ by Tony Van Eerd at Avast in Prague, that provides a great introduction to the concept of negative space in creative works, but it seems it is not available anymore.  Please watch this presentation, that it is in the same subject and still very good, but not necessarily directly applicable to this thread (I can't remember):
https://youtu.be/QTLn3goa3A8

benchmark/atoi-corpus.h
```diff
@@ -0,0 +1,135 @@
...
+    std::vector<int> skips_;
+    std::string characters_;
+
+    CorpusStringLength(std::vector<int> &&skips, std::string &&cs):
+        skips_{std::move(skips)}, characters_{std::move(cs)}
+    {}
+
+    template<typename G>
+    static auto makeCorpus(G &generator) {
+        auto count = 1031; // see Corpus8DecimalDigits for why 1031
```
Scottbruceheart, line 79: just make the 1031 a constant at this point, magic number rule

benchmark/atoi.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+// Copied from Daniel Lemire's GitHub at
+// https://lemire.me/blog/2018/10/03/quickly-parsing-eight-digits/
+// https://github.com/lemire/Code-used-on-Daniel-Lemire-s-blog/blob/ddb082981228f7256e9a4dbbf56fd4a335d78e30/2018/10/03/eightchartoi.c#L26C1-L34C2
+
+uint32_t parse_eight_digits_swar(const char *chars) {
+  uint64_t val;
+  memcpy(&val, chars, 8);
+  val = val - 0x3030303030303030;
+  uint64_t byte10plus   = ((val        * (1 + (0xa  <<  8))) >>  8) & 0x00FF00FF00FF00FF;
```
Scottbruceheart, line 18: I hope he got the PEMDAS right for real for real

benchmark/atoi-corpus.h
thecppzoo: The 9 arises from 8 bytes parsed and 1 null.  This is an artifact of my choice to use a function in the family of `printf` to convert from numbers to string, but I thought about it, and the 9 characters make things mostly unaligned, which I think is more representative of real uses.

benchmark/atoi.cpp
```diff
@@ -0,0 +1,219 @@
...
+namespace zoo {
+
+std::size_t c_strLength(const char *s) {
+    using S = swar::SWAR<8, std::size_t>;
+    constexpr auto
+        MSBs = S{S::MostSignificantBit},
+        Ones = S{S::LeastSignificantBit};
+    S bytes;
+    for(auto base = s;; base += 8) {
+        memcpy(&bytes.m_v, base, 8);
```
Scottbruceheart, line 64: Is the compiler smart enough to figure out that we don't need a memcpy against the offset of base, we just need the "Frame" of the SWAR dropped on top of those 8 bytes? I don't think it is, but it'd avoid a load.

benchmark/atoi.cpp
```diff
@@ -0,0 +1,219 @@
...
+        // subtracting one (if their value is greater than 0x80).
+        // This provides a way to detect the first null: It is the first lane
+        // in firstNullTurnsOnMSB that "flipped on" its MSB
+        auto cheapestInversionOfMSBs = ~bytes;
+        auto firstMSBsOnIsFirstNull =
+            firstNullTurnsOnMSB & cheapestInversionOfMSBs;
+        auto onlyMSBs = zoo::swar::convertToBooleanSWAR(firstMSBsOnIsFirstNull);
+        if(onlyMSBs) { // there is a null!
+            auto firstNullIndex = onlyMSBs.lsbIndex();
+            return firstNullIndex + (base - s);
```
Scottbruceheart, line 80: Obviously this is wildly unsafe and will happily run through memory, just like the normal strlen.

Scottbruceheart:
Did I overlook the tests for the added functions that act as evidence that the benchmarks are doing the right thing?

The GCC strlen is painfully hard to read.  I much prefer our methods.

SWAR atoi comment doesn't read easily to me (I had to go reread Lemire's).  Needs a rework.

If we're intending this as demo code, that directly means we want other SWE to read and understand and believe correctness and be impressed with as little cognitive effort on their part as possible.

benchmark/atoi.cpp
thecppzoo: This is not unsafe.
If by safety we are going to mean 'it guards against any potential mistake', we will never be able to deliver, all code would be unsafe.
Indeed this also requires a null, just like the specification of `strlen`:
https://en.cppreference.com/w/cpp/string/byte/strlen
I prefer to represent strings using a variable length encoding of the size of the string, but alas, this is a demo to compare things to see the benefits of a library.  I've always thought the null-terminated strings to be problematic.

thecppzoo:
> Did I overlook the tests for the added functions that act as evidence that the benchmarks are doing the right thing?
> 
Yes, as of this commit, in `catch2swar-demos.cpp`:
https://github.com/thecppzoo/zoo/pull/69/files#diff-286f2a6168f774ab1f7b54d49d8549b8c721a01199a226197f6a752573790674

> 
> SWAR atoi comment doesn't read easily to me (I had to go reread Lemire's). Needs a rework.
> 
> If we're intending this as demo code, that directly means we want other SWE to read and understand and believe correctness and be impressed with as little cognitive effort on their part as possible.

This is not advertising nor marketing.  These demos are the necessary complement to the code to avoid frustrations of not being able to see things being used and to provide us feedback to drive further development.

What is the comment that doesn't read easily?

thecppzoo:
> The GCC strlen is painfully hard to read. I much prefer our methods.
> 

Thanks, but your comment is entirely subjective.

What is the lesson to be learned from GCC's code, or from "our methods"?  GCC's code had a programming technique lurking in it that took me three attempts in three years to fully grok.  It has several "low-hanging-fruit" performance mistakes, and our code implicitly showcases the superiority of little-endian encoding for this type of task.

I think that this deserves a whole essay to discuss in proper detail and context, it would be helpful to know your opinion with full articulation of its points.

benchmark/atoi-corpus.h
Scottbruceheart: I generally like the 'this name is semantically void' convention E uses in this repo.

Scottbruceheart: Though I admit it was jarring through unfamiliarity at first.

benchmark/atoi.cpp
Scottbruceheart: It is as safe as strlen, no safer, no less safe.  Just commentary for color.
I also think null terminated strings are problematic.
We can ship something later that has a count.

PR #70, opened 2024-02-11
SWAR Demos on ARM

jamierpond:
ifdef some stuff to get the benchmarks running on my ARM Mac.


benchmark/atoi-corpus.h
```diff
@@ -119,17 +119,28 @@ struct CorpusStringLength {
     }
 };
 
+
+#if defined(__AVX2__)
+#define AVX2_STRLEN_CORPUS_X_LIST \
+  X(ZOO_AVX, zoo::avx2_strlen)
```
thecppzoo, line 125: Nit: please 4 space indentation?

Not just that I dislike inconsistent indentation, but that it "clogs the flow" of most programmers. I honestly don't know if it is my subjective dislike or the objective problem of accidental levels of indentation what bothers me.

Thanks to this presenation by Kevlin Henney, which I think is a true classic of our discipline (time marked to make it more relevant to this comment),
https://youtu.be/oyyFKHpzL0Q&t=2217
I learned the problem of the accidental levels of indentation, in this case in particular, using 4 spaces (or something equivalent, all around) is the only way to convey "these spaces only mean one level of detail deeper", the use of two spaces and 4 spaces in other places begs the question of whether the two spaces have their own alternative meaning, etc.

benchmark/atoi-corpus.h
```diff
@@ -119,17 +119,28 @@ struct CorpusStringLength {
...
+
 #define STRLEN_CORPUS_X_LIST \
     X(LIBC_STRLEN, strlen) \
     X(ZOO_STRLEN, zoo::c_strLength) \
     X(ZOO_NATURAL_STRLEN, zoo::c_strLength_natural) \
     X(ZOO_MANUAL_STRLEN, zoo::c_strLength_manualComparison) \
-    X(ZOO_AVX, zoo::avx2_strlen) \
-    X(GENERIC_GLIBC_STRLEN, STRLEN_old)
+    X(GENERIC_GLIBC_STRLEN, STRLEN_old) \
+    AVX2_STRLEN_CORPUS_X_LIST
```
thecppzoo, line 137: Truly great.

thecppzoo:
This is an excellent solution to discriminate away non portable things.
We could keep this PR as you made it, but please let us not use the normal macros to detect platforms such as `__AVX__` since this becomes an annoyance when we want to preprocess "lightly" the files for the purposes of using the compiler explorer, consolidating the code in a stand-alone file.
So, please make an adaptation so that the detection of platform specific things is done via a zoo-specific macro as in this file:
https://github.com/thecppzoo/zoo/blob/master/inc/zoo/pp/platform.h
, perhaps, `ZOO_CONFIGURED_TO_USE_AVX()`

benchmark/atoi-corpus.h
jamierpond: Thanks for this. Habits from working in a two space cpp codebase.

jamierpond:
@thecppzoo does this look good to you now?

inc/zoo/pp/platform.h
```diff
@@ -9,6 +9,8 @@
 
 #define MSVC_EMPTY_BASES
 
+#define ZOO_CONFIGURED_TO_USE_AVX defined(__AVX2__) // TODO bring this in from CMake
+
```
thecppzoo, line 13: Why did you choose to postpone the question of whether AVX2 is defined to every use of the macro instead of the more natural
#ifdef AVX/ #define ZOO_CONFIGURED/ #endif
?
I think I prefer that the conditional preprocessing at the point of use of ZOO_CONFIGURED... macros do not rely on any platform macro.  But I have not thought long about the question, hence I ask you why you made the choice

jamierpond: So you'd prefer to see:
```cpp
#ifdef __AVX2__
#define ZOO_CONFIGURED_TO_USE_AVX2
#endif
```
As opposed to the above? I think the effects of either will be more or less equivalent?

thecppzoo: Your question: yes.
The effects are not the same, see my question (which you did not answer explicitly), in my preferred conditional preprocessing the macro `__AVX2__` is never used again after preprocessing platform.h.  In the code you wrote, every use of the `ZOO...` macro will query the definition of `__AVX2__`, you've "postponed" the condition to every use.  My question was whether you have a reason to prefer this behavior, especially since this is not the natural behavior.
It is slightly safer, and therefore better, to do this:
```
#ifdef __AVX2__
#define ZOO_CONFIGURED_TO_USE_AVX() 1
#else
#define ZOO_CONFIGURED_TO_USE_AVX() 0
```
Another minor thing, in 2024 I am not going to bother with the nuance of AVX versus AVX2, whenever the code talks of AVX or I refer to AVX, mentally you should understand I mean both; in exactly the same way we say SSE to refer to all of the SSEs, to mean x86-64 128 bit SIMD.

thecppzoo: By the way, the zoo repository does not use `#pragma once` because I frequently override headers.  Imagine I'm doing some experiment in the Compiler Explorer and want to quickly enable or disable AVX:
I could override the platform header via:
`#define ZOO_PLATFORM_MACROS_H`, then do what I would want with each platform macro.
Your formulation will transform `ZOO_CONFIGURED_TO_USE_AVX` to the other macro I can't control in practice, the one provided by the system of `__AVX2__`

benchmark/catch2swar-demo.cpp
```diff
@@ -36,7 +36,9 @@ TEST_CASE("Atoi benchmarks", "[atoi][swar]") {
     REQUIRE(fromLIBC_STRLEN == fromZOO_NATURAL_STRLEN);
     REQUIRE(fromZOO_NATURAL_STRLEN == fromZOO_MANUAL_STRLEN);
     REQUIRE(fromGENERIC_GLIBC_STRLEN == fromZOO_NATURAL_STRLEN);
+#if ZOO_CONFIGURED_TO_USE_AVX
     REQUIRE(fromZOO_AVX == fromZOO_STRLEN);
+#endif
```
thecppzoo, line 41: I prefer to indent the conditional preprocessing with the code where it appears, it is a refinement, or more detail, with regards to the current level of indentation.  I don't know why colleagues overwhelmingly have the preference to flatten all conditional preprocessing.  This is especially nasty when the conditional preprocessing nests.

inc/zoo/pp/platform.h
jamierpond: I see, yes that makes total sense.

jamierpond: There is no deliberate reason to do it in the way I initially did. I defer to your proposed method.

jamierpond: Is there a reason to add the braces to make the macro look like a function?

thecppzoo: It is a very slight preference: it forces the macro to be defined.  Imagine you rely on `#ifdef` for your conditional compilation; you may inadvertently have compilation options that inconsistently define or not a particular macro, and it will apparently build (but inconsistently), which is frightening.  If you use a function style macro, not only it needs to be defined, but expand to a preprocessing-time true or false; so, you can use the function-style macro everywhere consistently and catch inconsistencies

jamierpond: That makes sense to me... however I can't find other examples in zoo where this happens. Would you please idiot check me on these changes? 

```diff
diff --git a/benchmark/atoi-corpus.h b/benchmark/atoi-corpus.h
index 783b6b1..525b761 100644
--- a/benchmark/atoi-corpus.h
+++ b/benchmark/atoi-corpus.h
@@ -120,7 +120,7 @@ struct CorpusStringLength {
 };
 
 
-#if ZOO_CONFIGURED_TO_USE_AVX
+#if ZOO_CONFIGURED_TO_USE_AVX()
 #define AVX2_STRLEN_CORPUS_X_LIST \
     X(ZOO_AVX, zoo::avx2_strlen)
 #else
diff --git a/benchmark/catch2swar-demo.cpp b/benchmark/catch2swar-demo.cpp
index 316163f..7e2e17a 100644
--- a/benchmark/catch2swar-demo.cpp
+++ b/benchmark/catch2swar-demo.cpp
@@ -36,7 +36,7 @@ TEST_CASE("Atoi benchmarks", "[atoi][swar]") {
     REQUIRE(fromLIBC_STRLEN == fromZOO_NATURAL_STRLEN);
     REQUIRE(fromZOO_NATURAL_STRLEN == fromZOO_MANUAL_STRLEN);
     REQUIRE(fromGENERIC_GLIBC_STRLEN == fromZOO_NATURAL_STRLEN);
-#if ZOO_CONFIGURED_TO_USE_AVX
+#if ZOO_CONFIGURED_TO_USE_AVX()
     REQUIRE(fromZOO_AVX == fromZOO_STRLEN);
 #endif
 
diff --git a/inc/zoo/pp/platform.h b/inc/zoo/pp/platform.h
index b219226..c068259 100644
--- a/inc/zoo/pp/platform.h
+++ b/inc/zoo/pp/platform.h
@@ -2,9 +2,9 @@
 #define ZOO_PLATFORM_MACROS_H
 
 #ifdef __AVX2__
-#define ZOO_CONFIGURED_TO_USE_AVX 1
+#define ZOO_CONFIGURED_TO_USE_AVX() 1
 #else
-#define ZOO_CONFIGURED_TO_USE_AVX 0
+#define ZOO_CONFIGURED_TO_USE_AVX() 0
 #endif
 
 #ifdef _MSC_VER
```

Results these errors. 
```
In file included from /Users/jamiepond/projects/zoo/benchmark/catch2swar-demo.cpp:2:
/Users/jamiepond/projects/zoo/benchmark/atoi-corpus.h:123:5: error: function-like macro 'ZOO_CONFIGURED_TO_USE_AVX' is not defined
#if ZOO_CONFIGURED_TO_USE_AVX()
    ^
/Users/jamiepond/projects/zoo/benchmark/catch2swar-demo.cpp:39:5: error: function-like macro 'ZOO_CONFIGURED_TO_USE_AVX' is not defined
#if ZOO_CONFIGURED_TO_USE_AVX()
    ^
2 errors generated.
make[2]: *** [CMakeFiles/catch2Benchmark.dir/catch2swar-demo.cpp.o] Error 1
make[1]: *** [CMakeFiles/catch2Benchmark.dir/all] Error 2
make: *** [all] Error 2
jamiepond@jamies-mac build % make  | pbcopy
In file included from /Users/jamiepond/projects/zoo/benchmark/catch2swar-demo.cpp:2:
/Users/jamiepond/projects/zoo/benchmark/atoi-corpus.h:123:5: error: function-like macro 'ZOO_CONFIGURED_TO_USE_AVX' is not defined
#if ZOO_CONFIGURED_TO_USE_AVX()
    ^
/Users/jamiepond/projects/zoo/benchmark/catch2swar-demo.cpp:39:5: error: function-like macro 'ZOO_CONFIGURED_TO_USE_AVX' is not defined
#if ZOO_CONFIGURED_TO_USE_AVX()
    ^
2 errors generated.
make[2]: *** [CMakeFiles/catch2Benchmark.dir/catch2swar-demo.cpp.o] Error 1
make[1]: *** [CMakeFiles/catch2Benchmark.dir/all] Error 2
make: *** [all] Error 2
```

If you could point us to other places zoo uses function-like macros to define a constant I would be grateful! Thanks!

thecppzoo: This would be the first use of function-like macros in this branch of the repository, inside Snap we introduced several.

I did not see where you #include the platform header.

jamierpond: I see, I assumed it was already included higher up since it ostensibly was already working. This already shows the value of the brackets! Thanks for this!!

benchmark/atoi-corpus.h
```diff
@@ -1,9 +1,9 @@
 #include "atoi.h"
-
 #include <vector>
 #include <string>
 #include <cstring>
 #include <random>
+#include <zoo/pp/platform.h>
```
thecppzoo, line 6: Micro-nit:

In this codebase, please always #include from the most specific to the more general, like this:
1. Headers specific to the feature related to the current file
2. General zoo headers (`platform.h` should probably be the last)
3. Specific external librareis
4. The standard library.

In this way, you give the chance to the local code to provide overrides for more general code.

benchmark/atoi-corpus.h
```diff
@@ -1,9 +1,9 @@
 #include "atoi.h"
-
 #include <vector>
```
thecppzoo, line 2:
```suggestion
#include "zoo/pp/platform.h"

#include <vector>
```

benchmark/atoi-corpus.h
```diff
@@ -1,9 +1,11 @@
 #include "atoi.h"
+#include "zoo/pp/platform.h"
 
 #include <vector>
 #include <string>
 #include <cstring>
 #include <random>
+#include <zoo/pp/platform.h>
```
thecppzoo, line 8:
```suggestion
```

benchmark/atoi-corpus.h
jamierpond: that makes sense !

benchmark/atoi-corpus.h
```diff
@@ -119,17 +119,28 @@ struct CorpusStringLength {
     }
 };
 
+
+#if ZOO_CONFIGURED_TO_USE_AVX()
+#define AVX2_STRLEN_CORPUS_X_LIST \
+    X(ZOO_AVX, zoo::avx2_strlen)
+#else
+#define AVX2_STRLEN_CORPUS_X_LIST /* nothing */
```
thecppzoo, line 127: I am going to immediately add cases to this X list, you'll see this was the proper way to encapsulate this.

Do you have "bandwidth" to help me do the equivalent for ARM? basically I can give you a diff patch that enables ARM, and hopefully it'll work, but if it doesn't you could fix it, of course that I will be online to help if you want

jamierpond: yup, i can add a patch if that helps. also feel free to push to this branch and i can pull and check it and fix errors i face

PR #75, opened 2024-02-27
Exponentiation WIP

jamierpond:
- some progress...
- hmmm
- add wip from fmtgp
- add notes
- add explicity typing to make compiler happy

### DIFF
--- a/inc/zoo/swar/associative_iteration.h
+++ b/inc/zoo/swar/associative_iteration.h
@@ -69,7 +69,7 @@ template<int NB, typename B>
 constexpr auto makeLaneMaskFromMSB(SWAR<NB, B> input) {
     using S = SWAR<NB, B>;
     auto msb = input & S{S::MostSignificantBit};
-    auto msbCopiedToLSB = S{msb.value() >> (NB - 1)};
+    auto msbCopiedToLSB = S{static_cast<B>(msb.value() >> (NB - 1))};
     return impl::makeLaneMaskFromMSB_and_LSB(msb, msbCopiedToLSB);
 }
 
@@ -239,10 +239,10 @@ constexpr auto multiplication_OverflowUnsafe_SpecificBitCount(
 
     auto halver = [](auto counts) {
         auto msbCleared = counts & ~S{S::MostSignificantBit};
-        return S{msbCleared.value() << 1};
+        return S{static_cast<T>(msbCleared.value() << 1)};
     };
 
-    multiplier = S{multiplier.value() << (NB - ActualBits)};
+    multiplier = S{static_cast<T>(multiplier.value() << (NB - ActualBits))};
     return associativeOperatorIterated_regressive(
         multiplicand, S{0}, multiplier, S{S::MostSignificantBit}, operation,
         ActualBits, halver
@@ -278,24 +278,24 @@ constexpr auto expo_OverflowUnsafe_SpecificBitCount(
     auto operation = [](auto left, auto right, auto counts) {
       const auto mask = makeLaneMaskFromMSB(counts);
       const auto antiMask = ~mask;
-
-      const auto product = multiplication_OverflowUnsafe_SpecificBitCount<ActualBits>(left, right);
+      const auto product =
+        multiplication_OverflowUnsafe_SpecificBitCount<ActualBits>(left, right);
       /*
        * if (count)
        *    return product;
-       *  else
+       * else
        *    return left;
-       * */
+       */
       return (product & mask) | (left & antiMask);
     };
 
     // halver should work same as multiplication... i think...
     auto halver = [](auto counts) {
         auto msbCleared = counts & ~S{S::MostSignificantBit};
-        return S{msbCleared.value() << 1};
+        return S{static_cast<T>(msbCleared.value() << 1)};
     };
 
-    exponent = S{exponent.value() << (NB - ActualBits)};
+    exponent = S{static_cast<T>(exponent.value() << (NB - ActualBits))};
     return associativeOperatorIterated_regressive(
         x, S{1}, exponent, S{S::MostSignificantBit}, operation,
         ActualBits, halver
--- a/inc/zoo/swar/associative_iteration.h
+++ b/inc/zoo/swar/associative_iteration.h
@@ -252,6 +252,7 @@ constexpr auto multiplication_OverflowUnsafe_SpecificBitCount(
 
 /*
      // extended from mathematics to generic programming
+     // see https://github.com/jamierpond/fmtgp/blob/main/2_first_algo/main.cpp
 
       template <typename T> constexpr T exp_acc(T r, T a, T n) {
         for (;;) {
--- a/inc/zoo/swar/associative_iteration.h
+++ b/inc/zoo/swar/associative_iteration.h
@@ -205,6 +205,27 @@ constexpr auto associativeOperatorIterated_regressive(
     return result;
 }
 
+template<int ActualBits, typename T>
+constexpr auto multiplication_scalar(
+    T multiplicand, T multiplier
+) {
+    auto operation = [](auto left, auto right, auto counts) {
+        return counts ? left + right : left;
+    };
+
+    auto halver = [](auto counts) {
+       return counts >> 1;
+    };
+
+    multiplier = multiplier + 1;
+
+    return associativeOperatorIterated_regressive(
+        multiplicand, 1, multiplier, 0, operation,
+        ActualBits, halver
+    );
+}
+
+
 template<int ActualBits, int NB, typename T>
 constexpr auto multiplication_OverflowUnsafe_SpecificBitCount(
     SWAR<NB, T> multiplicand, SWAR<NB, T> multiplier
@@ -224,6 +245,31 @@ constexpr auto multiplication_OverflowUnsafe_SpecificBitCount(
     multiplier = S{multiplier.value() << (NB - ActualBits)};
     return associativeOperatorIterated_regressive(
         multiplicand, S{0}, multiplier, S{S::MostSignificantBit}, operation,
+        ActualBits, halver
+    );
+}
+
+template<int ActualBits, int NB, typename T>
+constexpr auto expo_OverflowUnsafe_SpecificBitCount(
+    SWAR<NB, T> x,
+    SWAR<NB, T> exponent
+) {
+    using S = SWAR<NB, T>;
+
+    auto operation = [](auto left, auto right, auto counts) {
+      const auto product = multiplication_OverflowUnsafe_SpecificBitCount<ActualBits>(left, right);
+      const auto mask = makeLaneMaskFromMSB(counts);
+      return (mask & product) | (left & ~mask);
+    };
+
+    auto halver = [](auto counts) {
+        auto msbCleared = counts & ~S{S::MostSignificantBit};
+        return S{msbCleared.value() << 1};
+    };
+
+    exponent = S{exponent.value() << (NB - ActualBits)};
+    return associativeOperatorIterated_regressive(
+        x, S{1}, exponent, S{S::MostSignificantBit}, operation,
         ActualBits, halver
     );
 }
@@ -258,6 +304,17 @@ constexpr auto multiplication_OverflowUnsafe(
     return
         multiplication_OverflowUnsafe_SpecificBitCount<NB>(
             multiplicand, multiplier
+        );
+}
+
+template<int NB, typename T>
+constexpr auto expo_OverflowUnsafe(
+    SWAR<NB, T> base,
+    SWAR<NB, T> exponent
+) {
+    return
+       expo_OverflowUnsafe_SpecificBitCount<NB>(
+            base, exponent
         );
 }
 
--- a/inc/zoo/swar/associative_iteration.h
+++ b/inc/zoo/swar/associative_iteration.h
@@ -257,9 +257,18 @@ constexpr auto expo_OverflowUnsafe_SpecificBitCount(
     using S = SWAR<NB, T>;
 
     auto operation = [](auto left, auto right, auto counts) {
-      const auto product = multiplication_OverflowUnsafe_SpecificBitCount<ActualBits>(left, right);
       const auto mask = makeLaneMaskFromMSB(counts);
-      return (mask & product) | (left & ~mask);
+      const auto antiMask = ~mask;
+
+      const auto product = multiplication_OverflowUnsafe_SpecificBitCount<ActualBits>(left, right);
+      /*
+       * if (count)
+       *    return product;
+       *  else
+       *    return left;
+       * */
+      return (product & mask) | (left & antiMask);
+
     };
 
     auto halver = [](auto counts) {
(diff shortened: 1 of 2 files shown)


inc/zoo/swar/associative_iteration.h
```diff
@@ -218,16 +218,69 @@ constexpr auto multiplication_OverflowUnsafe_SpecificBitCount(
...
+
+    // halver should work same as multiplication... i think...
+    auto halver = [](auto counts) {
+        auto msbCleared = counts & ~S{S::MostSignificantBit};
+        return S{static_cast<T>(msbCleared.value() << 1)};
+    };
+
+    exponent = S{static_cast<T>(exponent.value() << (NB - ActualBits))};
+    return associativeOperatorIterated_regressive(
+        x, S{1}, exponent, S{S::MostSignificantBit}, operation,
```
DutchTulipCharter, line 279: Your neutral is, lane wise, { 0, ..., 1 }, so, the upper lanes are initialized to 0
You could `S{meta::BitmaskMaker<BaseType, 1, BitsPerLane>::value()}`

thecppzoo, line 279: Your neutral is, lane wise, { 0, ..., 1 }, so, the upper lanes are initialized to 0
You could `S{meta::BitmaskMaker<BaseType, 1, BitsPerLane>::value()}`

inc/zoo/swar/associative_iteration.h
```diff
@@ -218,16 +218,69 @@ constexpr auto multiplication_OverflowUnsafe_SpecificBitCount(
...
+       * else
+       *    return left;
+       */
+      return (product & mask) | (left & antiMask);
+    };
+
+    // halver should work same as multiplication... i think...
+    auto halver = [](auto counts) {
+        auto msbCleared = counts & ~S{S::MostSignificantBit};
+        return S{static_cast<T>(msbCleared.value() << 1)};
```
thecppzoo, line 274: Yes, `halver` does the same thing as multiplication.

inc/zoo/swar/associative_iteration.h
jamierpond: ooh thanks for that, that makes total sense!

thecppzoo: Possibly, but using `broadcast` is a complication because it relies on multiplication.
I have been slowly internalizing what needs to be done to support SIMD: no assumptions about the availability of whole-base-type multiplication.  In ARM NEON, x86-64 SSE of 128 bits, there is no such thing, but at most lane-wise 64 bit multiplication, same with AVX 256 bits.

BitmaskMaker can be upgraded to support longer types.
The problem is that SIMD negate `constexpr`, it's going to give us pain to make the non-constexpr flavor of the library

Scottbruceheart: That is, indeed, subtle.

Perhaps not so subtle in the SIMD `constexpr` negation...

test/swar/BasicOperations.cpp
```diff
@@ -41,6 +41,15 @@ static_assert(
     multiplication_OverflowUnsafe_SpecificBitCount<3>(Micand, Mplier).value()
 );
```
jamierpond, line 43: GOOD NEWS. Looks like my initial, first-pass implementation was correct after all!
Seems like part of the issue was how I'm trying to compare the binary literals.

All value I tried when evaluating using the hex idiom you guys have already got provides consistently correct answers.

inc/zoo/swar/associative_iteration.h
```diff
@@ -69,7 +69,7 @@ template<int NB, typename B>
 constexpr auto makeLaneMaskFromMSB(SWAR<NB, B> input) {
     using S = SWAR<NB, B>;
     auto msb = input & S{S::MostSignificantBit};
-    auto msbCopiedToLSB = S{msb.value() >> (NB - 1)};
+    auto msbCopiedToLSB = S{static_cast<B>(msb.value() >> (NB - 1))};
```
jamierpond, line 72: My compiler was complaining about needing casts, but I'm guessing there is probs a way to not need this?

inc/zoo/swar/associative_iteration.h
```diff
@@ -205,6 +210,34 @@ constexpr auto associativeOperatorIterated_regressive(
     return result;
 }
 
+
```
jamierpond, line 213: @thecppzoo It would be useful to get your feedback on this example. Been scratching my head to understand this, and I think I might need a nudge for it to click. I think it's the `forSquaring` which is still confusing to me. 

I think for me being able to generalise between a vanilla type and a SWAR type using the same template would really help me understand the concept more firmly.

inc/zoo/swar/associative_iteration.h
thecppzoo: No, the explicit casts seem required, I don't understand myself why

inc/zoo/swar/associative_iteration.h
```diff
@@ -205,6 +210,34 @@ constexpr auto associativeOperatorIterated_regressive(
...
+    auto operation = [](auto left, auto right, auto count) {
+      if (count) {
+        return left + right;
+      } else {
+        return left;
+      }
+    };
+
+    auto halver = [](auto count) {
+      return count >> 1;
```
thecppzoo, line 226: This is incorrect, this consumes the least significant bit, thus this is a halver for the "progressive" formulation, and you are using it for the "regressive".
The confusion needs me to apologize, I'm calling "halver" to something that in reality ought to double.  I'm trying to describe "the thing that messes with the counts in a way that determines what to do the next iteration"

test/swar/BasicOperations.cpp
```diff
@@ -41,6 +41,15 @@ static_assert(
...
 
+TEST_CASE("Jamie's totally working exponentiation :D") {
```
thecppzoo, line 44: Ah, thanks for the touch of humor, I think I will require good sense of humor in the style guide.  It cheers me to see the quality of the repository improving bit by bit.

inc/zoo/swar/associative_iteration.h
jamierpond: ahhh yes... of course... should have been more attentive to the other ones, and undestood that. thanks!

test/swar/BasicOperations.cpp
```diff
@@ -41,6 +41,15 @@ static_assert(
...
+    constexpr auto base     = SWAR<8, u32>{0x02'00'05'06}; //   2 | 0 |  5 |   6
+    constexpr auto exponent = SWAR<8, u32>{0x07'00'02'03}; //   7 | 0 |  2 |   3
```
thecppzoo, line 46: Among the purposes of tests is to help readers see things used, so they can mimic what they see until they develop more real understanding.
Let's not use confusing examples like 0^0 which is undefined.

.gitignore
```diff
@@ -5,3 +5,5 @@
 test/.vscode
 build
 .cache
+.idea
+**cmake-build**
```
thecppzoo, line 9: Does your IDE really require building inline with the sources, or within the source tree?
I am loathe to pollute the repository with support for the anti-pattern of building within the sources; however, Visual Studio Code forced me to yield.  VSCode is a poor tool that glitches in very frustrating ways when asked to build out of the source tree, that's the only reason for the `test/.vscode`, `build` and `.cache`.
So, unless your tool glitches, adopt the practice of not building in-tree and remove this.

inc/zoo/swar/associative_iteration.h
```diff
@@ -191,12 +191,17 @@ template<
...
+    const IterationCount forSquaring,
+    const Operator op,
+    unsigned log2Count,
+    const CountHalver ch
 ) {
     auto result = neutral;
-    if(!log2Count) { return result; }
-    for(;;) {
+    if (!log2Count) { return result; }
+    for (;;) {
```
thecppzoo, line 204: Please do not re-style code

inc/zoo/swar/associative_iteration.h
```diff
@@ -218,13 +251,57 @@ constexpr auto multiplication_OverflowUnsafe_SpecificBitCount(
...
+        multiplier,
+        S{S::MostSignificantBit},
+        operation,
+        ActualBits,
+        halver
+    );
+}
+
+template<int ActualBits, int NB, typename T>
+constexpr auto expo_OverflowUnsafe_SpecificBitCount(
```
thecppzoo, line 270: I must object to this identifier:
See that I am using unseemly long identifiers looking to make sure the names are clear, there is dissonance with using the `expo` abbreviation for exponentiation while the specification of the flavor of the operation takes 10 times as many characters.

jamierpond: Agreed.

inc/zoo/swar/associative_iteration.h
jamierpond: sorry, habit of fiddling while thinking.

thecppzoo:
Thanks for this effort.
Part of the value of re-expressing known techniques as what we call "associative iteration" is that this gets us closer to use any advantage we can find, later, about associative operations.
For example, we can identify cases in which the "counts" for the iteration are constants and device optimal addition chains for them.
Take cryptographic techniques: they may involve exponentiation (of some weird form of multiplication) by a constant: if it is a constant, chances are we could have compilation-time code that comes up with the optimal addition chain...

.gitignore
jamierpond: Why is this such a bad thing? I think most tools / LSPs expect to build in the source tree? I think CLion does this by default and I think `clangd` by default looks for `./build` in the root of the project?

thecppzoo: Just two things I dislike:
Most tools like the command line `grep`, `find`, etc. are extremely hard to configure for understanding `.gitignore`, so, you just slow down your work flow by having to treat the non-sources directories as explicit special cases with every non-git tool you use.

Let's say you want to build with GCC and also with Clang.
Who gets to build in-tree? both?
So, for every flavor of build you may want you need to provide an entry point within the tree, and add it to the .gitignore.
How's that helpful instead of segregating builds from sources?

These are two instances of the more general problem of disorganization.  A freshly cloned repository ought to contain only files generated by people or by tools that are so sophisticated they may as well.  Further processing of "sources" ought not to dump the results into a place where it is hard to tell them apart from files made by people.

It baffles me that colleagues expect that whatever configuration is good for holding source files is going to be just as good for built files.  For example, what if I have a configuration of very small blocks at the file system level expecting that source code is among the very small files in the distribution of file size, but keep the build filesystem with large blocks, because of the same reason on the opposite?

See? there are multiplicity of day to day advantages of keeping things organized.  Very fundamental stuff for Software Engineers.

test/swar/BasicOperations.cpp
```diff
@@ -41,6 +41,15 @@ static_assert(
     multiplication_OverflowUnsafe_SpecificBitCount<3>(Micand, Mplier).value()
 );
 
+TEST_CASE("Jamie's totally working exponentiation :D") {
+    constexpr auto base     = SWAR<8, u32>{0x02'01'05'06}; //   2 | 1 |  5 |   6
+    constexpr auto exponent = SWAR<8, u32>{0x07'00'02'03}; //   7 | 0 |  2 |   3
+    constexpr auto expected = SWAR<8, u32>{0x80'01'19'D8}; // 128 | 1 | 25 | 216
```
thecppzoo, line 47: Am I reviewing code made by ChatGPT?
It's technically correct to substitute the undefined 0^0 by the perfectly defined 1^0.  Still, aren't there more useful examples? Like 3^5 = 243?
Also, think about this suggestion:
`expected = SWAR<8, u32>{(127 << 24) + (1 << 16) + (25 << 8) + (216 << 0)}`
help us with some notation, macros may be considered, to initialize the lanes of a SWAR.
The explicit `meta::BitmaskMaker<....>::value` has been very tedious, error prone, time consuming for just repeating a constant, let alone initialize lanes to dissimilar values.

inc/zoo/swar/SWAR.h
```diff
@@ -70,6 +71,19 @@ struct SWAR {
 
     constexpr T value() const noexcept { return m_v; }
 
+   constexpr static T baseFromLaneLiterals(std::initializer_list<T> args) noexcept {
```
jamierpond, line 74: @thecppzoo i wonder what you think about this utility?

inc/zoo/swar/SWAR.h
```diff
@@ -70,6 +71,19 @@ struct SWAR {
...
+       T result = 0;
+       for (auto arg: args) {
+           result = (result << NBits) | arg;
+       }
+       return result;
```
thecppzoo, line 79: This is big endian, have you thought about which one ought to be the "canonical" endianness for literals?

inc/zoo/swar/SWAR.h
jamierpond: yeah i was trying a few variants with variadic arguments but that was a little tricky when trying to iterate over them, i feel like there is yet a better solution. 

this is one i was inspired for right now, but i think i will defo have more ideas with respect to ergonomics!

.gitignore
jamierpond: Sorry took a while to get back to this. 

I understand both of these points, I think the overriding point from me regardless of my opinion is that this is how it is going to be used by people, so we might as well make it convenient for people to contribute. Many IDEs and tools like the build folder to be in the source tree. I'm using the `clangd` LSP which is hooking into nvim which I believe requires the `compile_commands.json` to be in `./build`. JetBrains also will build in the source tree by default into `cmake-build-[BUILD_TYPE]`.  

Secondly, you can use tools like `ripgrep` in place of grep which are `.gitignore` sensitive, which is awesome.

Though I hear your points well and agree, this addition doesn't prevent more optimal workflows and enables anyone to contribute more easily.

inc/zoo/swar/SWAR.h
jamierpond: @thecppzoo how does this update look? makes it unambiguous that the number of lanes to use. 
how would you go about using the different types for the arguments here dynamically?

inc/zoo/swar/SWAR.h
jamierpond: I think it should read most significant on the left, least significant on the right, like it is now.

.gitignore
thecppzoo: Let's continue this conversation online, at this point, it is an item that should not go in this PR, but in another PR if we decide to support building in-tree.

Let's have PRs with a "single responsibility".

Scottbruceheart:
'overflow unsafe' points out the follow-up to this PR that exponentiation via repeated multiplication via repeated saturating addition is an implementation option.

thecppzoo:
> 'overflow unsafe' points out the follow-up to this PR that exponentiation via repeated multiplication via repeated saturating addition is an implementation option.

I don't know, Scott, perhaps it is better to focus on the flow of doubling the precision, saturating, and halving the precision.

In general, the "precision doubling" (casting to a pair of SWARs with double lane width, --how do we call this better than "precision doubling"?--) is a very useful technique, that we ought to illustrate often; see what I did with Lemire's conversion of ASCII to integer, see what the assembler of GLIBC did to implement `strlen` "masks".

PR #76, opened 2024-02-28
Atoi demo and parallel extraction (compress)

thecppzoo:
Indeed this PR demos `atoi` and parallel extraction.

During this PR I got to the conclusion that `operator not` for `BooleanSWAR` was a mistake:  `BooleanSWAR` must keep its "simdness", otherwise it collapsed into a boolean, then the use `not` for an array type was a categorical error.  This error leaked into Robin Hood but was identified and corrected early enough.


benchmark/atoi-corpus.h
```diff
@@ -143,10 +145,171 @@ struct CorpusStringLength {
...
+            return cp;
+        }
+
+        auto next() noexcept {
+            ++(*this);
+            return sentinel != skips;
+        }
+    };
+
+    Iterator commence() {
```
Scottbruceheart, line 207: why commence instead of begin or start? Seems like an unusual name, so probably warrants an explanatory comment.

thecppzoo: I hope you strongly agree with all of these reasons:

1. I'd like everybody to be able to express their ideas in the way most natural to them, without being encumbered with justifications.  In this way, we maximize the benefit of a plurality of ways to express things.  For example, you introduced "Lanes" as part of the nomenclature of SWAR, I was fastidiously talking of "sub elements", trying to avoid the confusion of people thinking the library used true support for lanes in the way of assembler specific instructions, or something like that, until "sub elements" began to feel insufferable, I adopted "Lanes" and I'm "running" with it.  Of course I was scared of something I should not have been scared by thinking people might get confused here.  Obviously, the whole point of the SWAR is to have "virtual" lanes, nobody will misunderstand this.
1.1. It is important to not require justifications.  Otherwise the default is "always talk in rhyme" as in Tony Van Eerd's presentation, style is not for the code to have a regularity without practical benefit, style is to facilitate the programmer to express ideas, hence it is subjective.  If you choose to indent to two spaces while most of the code is indented at four spaces, I will explain to you the reason against that it introduces accidental levels of indentation, something I avoid not just with the number of spaces, but semantically (I don't indent to the start of the arguments in a function call), the idea is that, in the absence of a superior recommendation, I find that making the indentation level being the same as the semantical nesting level (not syntactical) is helpful to comprehension.
2. In the practical sense, I don't think it helps productivity to explain everything that is unusual.  Otherwise, we'd have to start by having an idea of how unusual things must be before having to be justified.
3. By explaining trivial things explicitly, one devalues the value of important explanations.  A comment is an imposition on everyone: "I write this because I think it is necessary, thus, if you see my comments, read them", then you are forced to notice if the comment changed or if the code invalidated the comment during its evolution.
4. `begin` or `start` might evoke the wrong style of iteration, this is Java's.  I think `commence` is not used elsewhere, then it is perfect to denote "an iteration style that is not the usual style in C++".
5. `Iterator` is used because that's what it is, however, the use of `Iterator` is not important in this code because this type is supposed to be used as an opaque type by the user, all the knowledge needed is to see that termination is not determined, from the programmer's perspective, by the comparison against a sentinel, but the falseness of `next()`.

Could you articulate better why an unusual name may warrant an explanation?

inc/zoo/swar/SWAR.h
```diff
@@ -244,14 +245,20 @@ struct BooleanSWAR: SWAR<NBits, T> {
...
-        return BooleanSWAR(MaskLaneMSB ^ *this);
+    constexpr auto operator ~() const noexcept {
+        return BooleanSWAR(Base{Base::MostSignificantBit} ^ *this);
     }
 
+    #define BOOLEANSWAR_BINARY_LOGIC_OPERATOR X(^) X(&) X(|)
+    #define X(op) \
+        constexpr BooleanSWAR operator op(BooleanSWAR other) const noexcept { return this->Base::operator op(other); }
+    BOOLEANSWAR_BINARY_LOGIC_OPERATOR
+    #undef X
```
thecppzoo, line 256: use the canonical name `BOOLEANSWAR_BINARY_LOGIC_OPERATOR_X_LIST`

test/swar/BasicOperations.cpp
```diff
@@ -110,6 +110,17 @@ TEST_CASE("Compress/Expand", "[swar]") {
     using S1_32 = SWAR<32, uint32_t>;
     auto q = compress(S1_32{ToMove}, S1_32{Mask});
     CHECK(result == q.value());
+    SECTION("Regression 1") {
+        u64
+            input =   0b1010'1001'0110'0001'1001'0000'0010'1010'0100'0111'1110'1001'1111'0001'1110'1011,
+            mask =    0b0110'0000'0001'0101'0101'1111'0101'1100'0110'1111'0100'0111'0001'1000'0101'0010,
+            expected =0b0001'0000'0000'0001'0001'0000'0000'0010'0010'0111'0001'0001'0001'0000'0010'0001,
+            butGot =  0b0010'0000'0000'0001'0001'0000'0000'1000'0100'0111'0100'0001'0001'0000'0100'0010;
```
thecppzoo, line 118: not used

benchmark/atoi-corpus.h
Scottbruceheart: """begin or start might evoke the wrong style of iteration, this is Java's. I think commence is not used elsewhere, then it is perfect to denote "an iteration style that is not the usual style in C++".""" -> 'we use commence to denote an iteration style that isn't the usual C++ iteration style'.

Your points are valid, but a new to c++ or not c++ focused reader of this would certainly benefit from a short comment explicitly saying "This isn't the usualy C++ iterator!".  My request was only to benefit them.

Based on size of response, maybe you don't want to do it. That's ok too.

benchmark/atoi.cpp
```diff
@@ -115,6 +123,110 @@ blockAlignedLoad(PtrT *pointerInsideBlock, Block *b) {
...
+
+auto leadingDigitsCount(const char *p) noexcept {
+    using S = swar::SWAR<8, uint64_t>;
+    S bytes;
+    auto [base, misalignment] = blockAlignedLoad(p, &bytes.m_v);
+    auto bitDisplacement = 8 * misalignment;
+    constexpr static S
+        AllZeroCharacter{meta::BitmaskMaker<uint64_t, '0', 8>::value},
+        AllOn = ~S{0};
+    // blit the zero-characters to the misaligned part
```
Scottbruceheart, line 157: If we're going to shift something, and we're counting leading digits, why not shift out the misaligned part?  Load "xx12'34ab" , shift to "1234'abxx"?, get 4?

benchmark/atoi-corpus.h
thecppzoo: I still believe in point 3 above, it is counterproductive to explain the easy things.
The length of the response is to make clear that we want to allow people to express themselves and deal with the inconsistencies as readers and reviewers and maintainers.
I think the discussion is important, you mention two things: unusual things may warrant an explanation, we agree a short comment might benefit the new to C++ or not C++ focused reader, but how much do we have to cater to them?
My use of `commence` caught your attention, it did its job.
A common saying from Spanish: "A buen entendedor, pocas palabras", "to a good understander, few words".

benchmark/atoi.cpp
thecppzoo: The alternative you suggest is valid but sub-optimal, I actually used that option until I detected the improvement:
In this prolog to the loop, this code only brings the loop invariant in.
The loop invariant:
you work with the whole block, `base + nonDigitIndex - p` is the count
As far as I am able to tell, you'd need a modified version of the loop to calculate the count in a different way and hoist it to the prolog to work as an special first iteration, leading to non-optimal object code: in the code as it is, the extra blitting is better than a modified iteration.
Please check my reasoning, there might be a slight improvement here I may be missing.

Scottbruceheart: I think you are right.  We can increment base by the seen number of spaces in the prologue, but then we have to change the exit condition to something else (and I think that 'something else' is more expensive than 'did leading spaces returned eight' as a predicate).  Maybe there is some way, but everything looks like a modulo or a trick.)

inc/zoo/swar/SWAR.h
Scottbruceheart: are these notes to self?

Scottbruceheart (approved):
I'd prefer some explicit utests instead of 'does the output match glibc', but I get it.

thecppzoo:
> I'd prefer some explicit utests instead of 'does the output match glibc', but I get it.

It's how this code evolved: I just wanted to demo the library, I ended up having such performance that this became production code, and in several ways, best-of-its-kind; for a demo it would have sufficed to compare to the standard library.  You're right that for code we could recommend the world to use in preference over the standard library implementations, it ought to be unit tested more stringently; absolutely no bugs would be allowable.

This PR is to introduce these functions as a demo.  If we make a PR to refactor to the actual library, make sure to demand better unit tests

Scottbruceheart:
Doing it in a later PR is fine.

inc/zoo/swar/SWAR.h
```diff
@@ -244,14 +245,20 @@ struct BooleanSWAR: SWAR<NBits, T> {
...
-    constexpr auto operator not() const noexcept {
-        return BooleanSWAR(MaskLaneMSB ^ *this);
+    constexpr auto operator ~() const noexcept {
+        return BooleanSWAR(Base{Base::MostSignificantBit} ^ *this);
     }
 
+    #define BOOLEANSWAR_BINARY_LOGIC_OPERATOR X(^) X(&) X(|)
+    #define X(op) \
+        constexpr BooleanSWAR operator op(BooleanSWAR other) const noexcept { return this->Base::operator op(other); }
+    BOOLEANSWAR_BINARY_LOGIC_OPERATOR
```
thecppzoo, line 255:
```suggestion
    #define BOOLEANSWAR_BINARY_LOGIC_OPERATOR_X_LIST  X(^) X(&) X(|)
    #define X(op) \
        constexpr BooleanSWAR operator op(BooleanSWAR other) const noexcept { return this->Base::operator op(other); }
    BOOLEANSWAR_BINARY_LOGIC_OPERATOR_X_LIST
```

test/swar/BasicOperations.cpp
thecppzoo, line 118:
```suggestion
```

Scottbruceheart:
Performed the merge against main, submitting.

Future hardening work needs test for be intrastructure level.

PR #78, opened 2024-03-22
Implement GreaterLess with MSB on for SWAR.

Scottbruceheart:
Includes testing that exercises MSB on and various call patterns. Could be stronger.

TAOCP claims the formula used takes 8 ops.

### DIFF
--- a/inc/zoo/swar/SWAR.h
+++ b/inc/zoo/swar/SWAR.h
@@ -394,7 +394,7 @@ constexpr T
 indexOfMostSignficantLaneSet(SWAR<NBits, T> test) noexcept {
     const auto TypeWidth = sizeof(T) * 8;
     const auto TopVal = (T{1}<<(TypeWidth-NBits))-1, BottomVal = (T{1}<<(NBits-1))-1;
-    const u64 MappingConstant = TopVal / BottomVal;
+    const T MappingConstant = TopVal / BottomVal;
     return (test.value() * MappingConstant) >> (TypeWidth - NBits);
 }
 
--- a/inc/zoo/swar/SWAR.h
+++ b/inc/zoo/swar/SWAR.h
@@ -188,7 +188,7 @@ constexpr auto isolateLSB(T v) {
 
 template<int NBits, typename T>
 constexpr auto leastNBitsMask() {
-    return (T(1ull)<<NBits)-1;
+    return (T{1}<<NBits)-1;
 }
 
 template<int NBits, uint64_t T>
@@ -198,7 +198,7 @@ constexpr auto leastNBitsMask() {
 
 template<int NBits, typename T = uint64_t>
 constexpr T mostNBitsMask() {
-  return ~leastNBitsMask<sizeof(T)*8-NBits, T>();
+    return ~leastNBitsMask<sizeof(T)*8-NBits, T>();
 }
 
 
@@ -298,7 +298,7 @@ struct BooleanSWAR: SWAR<NBits, T> {
 
     template<int NB, typename TT>
     constexpr T
-    fastFirstLeftOn(SWAR<NBits, T> test) noexcept;
+    indexOfMostSignficantLaneSet(SWAR<NBits, T> test) noexcept;
 
     template<int NB, typename TT>
     friend constexpr BooleanSWAR<NB, TT>
@@ -367,7 +367,7 @@ constantIsGreaterEqual_MSB_off(SWAR<NBits, T> subtrahend) noexcept {
 
 template<typename T, typename U, typename V>
 constexpr T median(T x, U y, V z) {
-  return (x | y) & (y | z) & (x | z);
+    return (x | y) & (y | z) & (x | z);
 }
 
 template<int NBits, typename T>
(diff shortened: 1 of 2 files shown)


inc/zoo/swar/SWAR.h
```diff
@@ -362,6 +365,39 @@ constantIsGreaterEqual_MSB_off(SWAR<NBits, T> subtrahend) noexcept {
     }
 }
 
+template<typename T, typename U, typename V>
+constexpr T median(T x, U y, V z) {
+  return (x | y) & (y | z) & (x | z);
```
thecppzoo, line 370: Accidental level of indentation

inc/zoo/swar/SWAR.h
```diff
@@ -362,6 +365,39 @@ constantIsGreaterEqual_MSB_off(SWAR<NBits, T> subtrahend) noexcept {
...
+}
+
+// In the condition where only MSBs will be on, we can fast lookup with 1 multiply the index of the leftmost byte.
+// This appears to be a mapping from the (say) 256 unique values of a 64 bit int where only MSBs of each 8 bits can be on, but I don't fully understand it.
+// Adapted from TAOCP Vol 4A Page 153 Eq 94.
+template<int NBits, typename T>
+constexpr T
+fastFirstLeftOn(SWAR<NBits, T> test) noexcept {
+    const auto width = sizeof(T) * 8;
+    const auto tval = (1ull<<(width-NBits))-1, bval = (1ull<<(NBits-1))-1;
```
thecppzoo, line 396: Let us not initialize constants to 64 bit `unsigned`s, as in `1ull`, it is a bug lurking when we use anything other than integers as the base type of `SWAR`, we should be very watchful to initialize these constants to `T{1}` for example.
Also, it may be performance disadvantageous if the base type is not 64 bits, since perhaps the use of `uint64_t` here might trigger a promotion to 64 bits and a demotion to whatever `T` is.

thecppzoo: Also, please give it a try to naming the compile time constants with `BactrianCase` instead of `dromedaryCase`

test/swar/BasicOperations.cpp
```diff
@@ -299,6 +307,64 @@ GE_MSB_TEST(0x7777'7777,
...
+GE_MSB_ON_TEST(0x0123'4567,
+               0x7777'7777,
+               0x0000'0008)
+GE_MSB_ON_TEST(0x0000'0000,
+               0x0123'4567,
+               0x8000'0000)
+GE_MSB_ON_TEST(0x7777'7777,
+               0x0123'4567,
+               0x8888'8888)
+
```
thecppzoo, line 337: Please let us not use accidental indentation levels.

test/swar/BasicOperations.cpp
```diff
@@ -299,6 +307,64 @@ GE_MSB_TEST(0x7777'7777,
...
+TEST_CASE(
+    "greaterEqualMSBOn",
+    "[swar][signed-swar][unsigned-swar]"
```
thecppzoo, line 340: Wait, does this implementation of greaterEqual is aware of two's complement?
Why does this have the tag `signed-swar`?

inc/zoo/swar/SWAR.h
Scottbruceheart: One day I will fix my habit, apologies.

inc/zoo/swar/SWAR.h
Scottbruceheart: Good catch, I got sloppy on the constant inits.

inc/zoo/swar/SWAR.h
```diff
@@ -362,6 +365,39 @@ constantIsGreaterEqual_MSB_off(SWAR<NBits, T> subtrahend) noexcept {
...
+    // z = (x|h) - (y&~h)
+    using S = swar::SWAR<NBits, T>;
+    const auto h = S::MostSignificantBit, x = left.value(), y = right.value();  // x=left, y= right is x < y
+    const auto z = (x|h) - (y&~h);
+    // bitwise ternary median! 
+    const auto t = h & ~median(x, ~y, z);
+    return ~BooleanSWAR<NBits, T>{static_cast<T>(t)};  // ~(x<y) === x >= y
+}
+
+// In the condition where only MSBs will be on, we can fast lookup with 1 multiply the index of the leftmost byte.
```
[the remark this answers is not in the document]
Scottbruceheart: It is an intentionally terrible name (Knuth uses lambda and rho), and the intended use is so I can throw a less than and then know which lane failed PSL check first... 😄 
I don't think I like calling it log, but let me contemplate it some more.

test/swar/BasicOperations.cpp
Scottbruceheart: They aren't accidental?  There is a comment at the top of the block on why they are set up that way, to be easily read.

test/swar/BasicOperations.cpp
Scottbruceheart: That's leftover from a previous mis-reading of either that Lamport paper or the Knuth book.  I'm not sure there is a way to that easily (though I'm going to try it at some point, no doubt)

inc/zoo/swar/SWAR.h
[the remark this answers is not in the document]
Scottbruceheart: I think the constants given map 1:1 a pattern like "100'100'100'100" (MSB only on in all lanes or nothing on in lane) to [0,2^(NumLanes)) but I don't yet understand it enough to test and use the generalization.
Also, throwing to a table lookup feels like it will only be useful for smaller SWAR ops.

For some reason I thought we already had SWAR logarithm somewhere?

test/swar/BasicOperations.cpp
Scottbruceheart: I'm fine with using standards: How do you suggest making these tests readable at a glance?  Our previous agreement was to use the nonstandard levels so they'd be readable.
```
GE_MSB_ON_TEST(
    0x7777'7777,
    0x0123'4567,
    0x8888'8888)
```
?

inc/zoo/swar/SWAR.h
```diff
@@ -386,16 +386,16 @@ greaterEqual(SWAR<NBits, T> left, SWAR<NBits, T> right) noexcept {
...
 constexpr T
-fastFirstLeftOn(SWAR<NBits, T> test) noexcept {
-    const auto width = sizeof(T) * 8;
-    const auto tval = (1ull<<(width-NBits))-1, bval = (1ull<<(NBits-1))-1;
-    const u64 cval = tval / bval; 
-    return (test.value() * cval) >> (width - NBits);
+indexOfMostSignficantLaneSet(SWAR<NBits, T> test) noexcept {
+    const auto TypeWidth = sizeof(T) * 8;
+    const auto TopVal = (T{1}<<(TypeWidth-NBits))-1, BottomVal = (T{1}<<(NBits-1))-1;
+    const u64 MappingConstant = TopVal / BottomVal;
```
thecppzoo, line 397: This explicit use of u64 is incorrect, `TopVal / BottomVal` is executed as a `T` divided by a `T`.  Why do you need to convert that result to u64?

PR #79, opened 2024-03-25
add fromLaneLiterals and tests

jamierpond:


inc/zoo/swar/SWAR.h
```diff
@@ -77,6 +77,21 @@ struct SWAR {
 
     constexpr T value() const noexcept { return m_v; }
 
+    template<std::size_t N>
+    constexpr static T baseFromLaneLiterals(const T(&args)[N]) {
+        static_assert(N == Lanes, "Wrong number of lanes");
```
thecppzoo, line 82: SFINAE this.
`std::enable_if_t<N == Lanes, T>`
Also, please separate the type from the identifier being declared as in `const T (&args)[`, in this codebase we bind the pointer and reference indicators to the identifier they apply to, not to the type they don't apply to, as is far more common.

inc/zoo/swar/SWAR.h
```diff
@@ -77,18 +77,17 @@ struct SWAR {
 
     constexpr T value() const noexcept { return m_v; }
 
-    template<std::size_t N>
-    constexpr static T baseFromLaneLiterals(const T(&args)[N]) {
-        static_assert(N == Lanes, "Wrong number of lanes");
-        T result = 0;
+    template<std::size_t N, typename = std::enable_if_t<N == Lanes, T>>
```
thecppzoo, line 80: No.
You changed the arity of the template.
```suggestion
template<std::size_t N>
std::enable_if<N == Lanes, T>
baseFromLaneLiterals(const T (&args)[N]) {
```

thecppzoo: forgot the `constexpr static`

test/swar/BasicOperations.cpp
```diff
@@ -26,6 +26,15 @@ using S32_32 = SWAR<32, uint32_t>;
...
 using S64_64 = SWAR<64, uint64_t>;
 
+static_assert(SWAR<8, u32>::fromLaneLiterals({0, 0, 0, 0}).value() == 0);
+static_assert(SWAR<8, u32>::fromLaneLiterals({0, 0, 0, 1}).value() == 1);
+static_assert(SWAR<8, u32>::fromLaneLiterals({8, 3, 2, 1}).value() == 0x08'03'02'01);
+static_assert(SWAR<8, u32>::fromLaneLiterals({42, 42, 42, 42}).value() == 0x2A'2A'2A'2A);
+
+static_assert(SWAR<4, u32>::fromLaneLiterals({0, 0, 0, 0, 0, 0, 0, 0}).value() == 0);
+static_assert(SWAR<4, u32>::fromLaneLiterals({0, 0, 0, 0, 0, 0, 0, 1}).value() == 1);
+static_assert(SWAR<4, u16>::fromLaneLiterals({8, 3, 2, 1}).value() == 0b1000'0011'0010'0001);
```
Scottbruceheart, line 36: In practice, people are going to set up something like S4_32::FLL({1,2,3,4})?

inc/zoo/swar/SWAR.h
```diff
@@ -77,6 +77,20 @@ struct SWAR {
 
     constexpr T value() const noexcept { return m_v; }
 
+    template<std::size_t N, typename = std::enable_if_t<N == Lanes, T>>
+    constexpr static T baseFromLaneLiterals(const T (&args)[N]) {
```
Scottbruceheart, line 81: Raises the follow-up: Do you think `BooleanSWAR(4,u16)::fromLaneLiterals(true, false, true, fasel)` would be useful?  I acknowledge the point you made the other day that most engineers aren't and will not be fluent in hexadecimal.

thecppzoo: Yes, let's do it, I'm "convinced" now:
```
SWAR my4x16{Literals<16, uint64_t>, {0xFF00, 0xF0F0, 0xFFFF, 0xF000}};
BooleanSWAR{Literals<16, uint64_t>, {true, true, false, true}};
```

inc/zoo/swar/SWAR.h
```diff
@@ -77,6 +77,26 @@ struct SWAR {
 
     constexpr T value() const noexcept { return m_v; }
 
+    // is this the right name?
+    constexpr static T maxLaneValue = (T{1} << NBits) - 1;
```
thecppzoo, line 81: Never shift by `NBits`, if the SWAR is "degenerated", in the sense that it contains only one lane, then you would be shifting all bits, the result is undefined.  This gave Scott and I a problem preparing for CPPCon 2022, it turns out that even between processors of the same brand, the result of shifting by the whole bit count is different, sometimes it is zero, some others does nothing.
Here's what you want:

```suggestion
    constexpr static MaxUnsignedLaneValue = {}() { // a lambda to give names to the intermediate steps
        T
            numberOneInTheLeastSignificantLane{1},
            turnedToMSB = numberOneInTheLeastSignificantLane << (NBits - 1),
            leastSignificantLaneFlipped = turnedToMSB - 1,
            rv = turnedToMSB | leastSignificantLaneFlipped;
        return rv;
    }();        
```
I did not make the rules.  This is exceedingly strange, that something as simple as this requires something so complicated.
Alternatively you can `~(((~T{0}) << (NBits - 1)) << 1)` or whatever else you want, but what you wrote is wrong, and in practice would lead you to get corrupt code silently.

inc/zoo/swar/SWAR.h
```diff
@@ -250,14 +270,25 @@ struct BooleanSWAR: SWAR<NBits, T> {
         return BooleanSWAR(swar::clearLSB(this->value()));
     }
 
+    template<std::size_t N, typename = std::enable_if_t<N == SWAR<NBits, T>::Lanes, T>>
+    constexpr static BooleanSWAR fromBooleanLiterals(const bool (&args)[N]) {
```
thecppzoo, line 274: Come on, don't do this, I find it very ugly to change the arity of a template to apply SFINAE

test/swar/BasicOperations.cpp
jamierpond: at least i think this makes it more fun and easy to write tests and to reason about correctness of the code.

inc/zoo/swar/SWAR.h
jamierpond: I see, check that out.

inc/zoo/swar/SWAR.h
jamierpond: ok i follow.

inc/zoo/swar/SWAR.h
jamierpond: ok will try and get this going!

inc/zoo/swar/SWAR.h
thecppzoo: (y)

inc/zoo/swar/SWAR.h
```diff
@@ -77,6 +77,25 @@ struct SWAR {
 
     constexpr T value() const noexcept { return m_v; }
 
+    // is this the right name?
+    constexpr static T MaxUnsignedLaneValue = ~(((~T{0}) << (NBits - 1)) << 1);
```
thecppzoo, line 81: in the standard librry, these things are called `UINT_MAX` and the like, they're macros.
We're not going to do that.
I much rather:
1. Have a descriptive name like "maximum value for an unsigned lane"
2. Make it a `T`, not a macro

This name satisfies the conditions, what's your potential objection to ask?

inc/zoo/swar/SWAR.h
```diff
@@ -4,389 +4,453 @@
...
 #endif
 
-namespace zoo { namespace swar {
+namespace zoo {
+namespace swar {
+
+template <int NBits, typename T> struct SWAR;
+
+template <int NumBits, typename BaseType> struct Literals_t {
+  constexpr static void (SWAR<NumBits, BaseType>::*value)() = nullptr;
```
jamierpond, line 20: @thecppzoo do we still want to keep this? Looks like we don't need it?

jamierpond:
Closing in favour of cleaner new PR.

PR #80, opened 2024-04-16
New Literals Utils

jamierpond:
moved this into a new PR for clarity. Literals can now be written like so:
```cpp
static_assert(SWAR{Literals<32, u64>, {2, 1}}.value() == 0x00000002'00000001);
static_assert(SWAR{Literals<32, u64>, {1, 2}}.value() == 0x00000001'00000002);
static_assert(SWAR{Literals<16, u64>, {4, 3, 2, 1}}.value() == 0x0004'0003'0002'0001);
static_assert(SWAR{Literals<16, u64>, {1, 2, 3, 4}}.value() == 0x0001'0002'0003'0004);
```

### DIFF
--- a/inc/zoo/swar/SWAR.h
+++ b/inc/zoo/swar/SWAR.h
@@ -82,7 +82,7 @@ struct SWAR {
         MaxUnsignedLaneValue = LeastSignificantLaneMask;
 
     template <typename U>
-    constexpr auto loadIntoLanes(const U (&values)[Lanes]) const noexcept {
+    constexpr static auto from_array(const U (&values)[Lanes]) noexcept {
         auto result = T{0};
         for (auto value : values) {
             result = (result << NBits) | value;
@@ -92,7 +92,7 @@ struct SWAR {
 
     template <typename Arg, std::size_t N, typename = std::enable_if_t<N == Lanes, int>>
     constexpr
-    SWAR(Literals_t<NBits, T>, const Arg (&values)[N]) : m_v{loadIntoLanes(values)} {}
+    SWAR(Literals_t<NBits, T>, const Arg (&values)[N]) : m_v{from_array(values)} {}
 
     SWAR() = default;
     constexpr explicit SWAR(T v): m_v(v) {}
--- a/inc/zoo/swar/SWAR.h
+++ b/inc/zoo/swar/SWAR.h
@@ -82,8 +82,8 @@ struct SWAR {
         MaxUnsignedLaneValue = LeastSignificantLaneMask;
 
     template <typename U, typename ManipulationFn>
-    constexpr auto loadBaseTypeIntoLanes(const U (&values)[Lanes],
-                                         const ManipulationFn&& manipulation) {
+    constexpr auto loadIntoLanes(const U (&values)[Lanes],
+                                 const ManipulationFn&& manipulation) {
         auto result = T{0};
         for (auto value : values) {
             auto laneValue = manipulation(value);
@@ -95,7 +95,7 @@ struct SWAR {
     template <typename Arg, std::size_t N, typename = std::enable_if_t<N == Lanes, int>>
     constexpr
     SWAR(Literals_t<NBits, T>, const Arg (&values)[N]) : m_v{0} {
-        m_v = loadBaseTypeIntoLanes(values, [](auto x) { return x; });
+        m_v = loadIntoLanes(values, [](auto x) { return x; });
     }
 
 
@@ -266,7 +266,7 @@ struct BooleanSWAR: SWAR<NBits, T> {
     template <std::size_t N, typename = std::enable_if_t<N == Base::Lanes, T>>
     constexpr BooleanSWAR(Literals_t<NBits, T>, const bool (&values)[N]) : Base{0} {
         constexpr auto msbOfFirstLane = T{1} << (NBits - 1);
-        this->m_v = Base::loadBaseTypeIntoLanes(values, [](auto x) { return x ? msbOfFirstLane : 0; });
+        this->m_v = Base::loadIntoLanes(values, [](auto x) { return x ? msbOfFirstLane : 0; });
     }
 
     // Booleanness is stored in the MSBs
--- a/inc/zoo/swar/SWAR.h
+++ b/inc/zoo/swar/SWAR.h
@@ -77,7 +77,8 @@ struct SWAR {
                 ~(~T(0) << NBits),
         // Use LowerBits in favor of ~MostSignificantBit to not pollute
         // "don't care" bits when non-power-of-two bit lane sizes are supported
-        LowerBits = MostSignificantBit - LeastSignificantBit;
+        LowerBits = MostSignificantBit - LeastSignificantBit,
+        MaxUnsignedLaneValue = LeastSignificantLaneMask;
 
     template <typename Arg, std::size_t N, typename = std::enable_if_t<N == Lanes, int>>
     constexpr
@@ -89,7 +90,6 @@ struct SWAR {
         m_v = result;
     }
 
-    constexpr static T MaxUnsignedLaneValue = ~(((~T{0}) << (NBits - 1)) << 1);
 
     SWAR() = default;
     constexpr explicit SWAR(T v): m_v(v) {}
--- a/inc/zoo/swar/SWAR.h
+++ b/inc/zoo/swar/SWAR.h
@@ -81,12 +81,11 @@ struct SWAR {
         LowerBits = MostSignificantBit - LeastSignificantBit,
         MaxUnsignedLaneValue = LeastSignificantLaneMask;
 
-    template <typename U, typename ManipulationFn>
-    constexpr auto loadIntoLanes(const U (&values)[Lanes], const ManipulationFn&& manipulation) {
+    template <typename U>
+    constexpr auto loadIntoLanes(const U (&values)[Lanes]) const noexcept {
         auto result = T{0};
         for (auto value : values) {
-            auto laneValue = manipulation(value);
-            result = (result << NBits) | laneValue;
+            result = (result << NBits) | value;
         }
         return result;
     }
@@ -94,7 +93,7 @@ struct SWAR {
     template <typename Arg, std::size_t N, typename = std::enable_if_t<N == Lanes, int>>
     constexpr
     SWAR(Literals_t<NBits, T>, const Arg (&values)[N])
-      : m_v{loadIntoLanes(values, [](auto x) { return x; })} {}
+      : m_v{loadIntoLanes(values)} {}
 
     SWAR() = default;
     constexpr explicit SWAR(T v): m_v(v) {}
@@ -261,10 +260,8 @@ struct BooleanSWAR: SWAR<NBits, T> {
     using Base = SWAR<NBits, T>;
 
     template <std::size_t N>
-    constexpr BooleanSWAR(Literals_t<NBits, T>, const bool (&values)[N]) : Base{0} {
-        constexpr auto msbOfFirstLane = T{1} << (NBits - 1);
-        this->m_v = Base::loadIntoLanes(values, [](auto x) { return x ? msbOfFirstLane : 0; });
-    }
+    constexpr BooleanSWAR(Literals_t<NBits, T>, const bool (&values)[N])
+    : Base(Literals<NBits, T>, values) { this->m_v <<= (NBits - 1); }
 
     // Booleanness is stored in the MSBs
     static constexpr auto MaskMSB =
--- a/inc/zoo/swar/SWAR.h
+++ b/inc/zoo/swar/SWAR.h
@@ -261,8 +261,10 @@ struct BooleanSWAR: SWAR<NBits, T> {
     using Base = SWAR<NBits, T>;
 
     template <std::size_t N>
-    constexpr BooleanSWAR(Literals_t<NBits, T>, const bool (&values)[N])
-       : Base(Literals<NBits, T>, values) { this->m_v << (NBits - 1); }
+    constexpr BooleanSWAR(Literals_t<NBits, T>, const bool (&values)[N]) : Base{0} {
+        constexpr auto msbOfFirstLane = T{1} << (NBits - 1);
+        this->m_v = Base::loadIntoLanes(values, [](auto x) { return x ? msbOfFirstLane : 0; });
+    }
 
     // Booleanness is stored in the MSBs
     static constexpr auto MaskMSB =
--- a/inc/zoo/swar/SWAR.h
+++ b/inc/zoo/swar/SWAR.h
@@ -82,8 +82,7 @@ struct SWAR {
         MaxUnsignedLaneValue = LeastSignificantLaneMask;
 
     template <typename U, typename ManipulationFn>
-    constexpr auto loadIntoLanes(const U (&values)[Lanes],
-                                 const ManipulationFn&& manipulation) {
+    constexpr auto loadIntoLanes(const U (&values)[Lanes], const ManipulationFn&& manipulation) {
         auto result = T{0};
         for (auto value : values) {
             auto laneValue = manipulation(value);
@@ -94,10 +93,8 @@ struct SWAR {
 
     template <typename Arg, std::size_t N, typename = std::enable_if_t<N == Lanes, int>>
     constexpr
-    SWAR(Literals_t<NBits, T>, const Arg (&values)[N]) : m_v{0} {
-        m_v = loadIntoLanes(values, [](auto x) { return x; });
-    }
-
+    SWAR(Literals_t<NBits, T>, const Arg (&values)[N])
+      : m_v{loadIntoLanes(values, [](auto x) { return x; })} {}
 
     SWAR() = default;
     constexpr explicit SWAR(T v): m_v(v) {}
@@ -263,11 +260,9 @@ template<int NBits, typename T>
 struct BooleanSWAR: SWAR<NBits, T> {
     using Base = SWAR<NBits, T>;
 
-    template <std::size_t N, typename = std::enable_if_t<N == Base::Lanes, T>>
-    constexpr BooleanSWAR(Literals_t<NBits, T>, const bool (&values)[N]) : Base{0} {
-        constexpr auto msbOfFirstLane = T{1} << (NBits - 1);
-        this->m_v = Base::loadIntoLanes(values, [](auto x) { return x ? msbOfFirstLane : 0; });
-    }
+    template <std::size_t N>
+    constexpr BooleanSWAR(Literals_t<NBits, T>, const bool (&values)[N])
+       : Base(Literals<NBits, T>, values) { this->m_v << (NBits - 1); }
 
     // Booleanness is stored in the MSBs
     static constexpr auto MaskMSB =
--- a/inc/zoo/swar/SWAR.h
+++ b/inc/zoo/swar/SWAR.h
@@ -12,7 +12,8 @@
 
 namespace zoo { namespace swar {
 
-template <int NBits, typename T> struct SWAR;
+template <int NBits, typename T>
+struct SWAR;
 
 template <int NumBits, typename BaseType> struct Literals_t {
   constexpr static void (SWAR<NumBits, BaseType>::*value)() = nullptr;
@@ -80,14 +81,21 @@ struct SWAR {
         LowerBits = MostSignificantBit - LeastSignificantBit,
         MaxUnsignedLaneValue = LeastSignificantLaneMask;
 
+    template <typename U, typename ManipulationFn>
+    constexpr auto loadBaseTypeIntoLanes(const U (&values)[Lanes],
+                                         const ManipulationFn&& manipulation) {
+        auto result = T{0};
+        for (auto value : values) {
+            auto laneValue = manipulation(value);
+            result = (result << NBits) | laneValue;
+        }
+        return result;
+    }
+
     template <typename Arg, std::size_t N, typename = std::enable_if_t<N == Lanes, int>>
     constexpr
     SWAR(Literals_t<NBits, T>, const Arg (&values)[N]) : m_v{0} {
-        auto result = T{0};
-        for (const auto arg : values) {
-            result = (result << NBits) | arg;
-        }
-        m_v = result;
+        m_v = loadBaseTypeIntoLanes(values, [](auto x) { return x; });
     }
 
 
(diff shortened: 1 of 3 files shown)


inc/zoo/swar/SWAR.h
```diff
@@ -12,6 +12,15 @@
 
 namespace zoo { namespace swar {
 
+template <int NBits, typename T> struct SWAR;
+
+template <int NumBits, typename BaseType> struct Literals_t {
+  constexpr static void (SWAR<NumBits, BaseType>::*value)() = nullptr;
```
thecppzoo, line 18: This is not in use, remove

inc/zoo/swar/SWAR.h
```diff
@@ -70,6 +79,18 @@ struct SWAR {
...
+    constexpr
+    SWAR(Literals_t<NBits, T>, const Arg (&values)[N]) : m_v{0} {
+        auto result = T{0};
+        for (const auto arg : values) {
+            result = (result << NBits) | arg;
+        }
+        m_v = result;
+    }
+
+    constexpr static T MaxUnsignedLaneValue = ~(((~T{0}) << (NBits - 1)) << 1);
```
thecppzoo, line 92: I agree this constant is worth including, but not here, and not with this calculation.  See there are a bunch of constants in `SWAR`, you may, for example, put it at the end of the list like this:
```c++
MaxUnsignedLaneValue = LeastSignificantLaneMask;
```

test/swar/BasicOperations.cpp
```diff
@@ -33,6 +33,26 @@ using S32_32 = SWAR<32, uint32_t>;
 
 using S64_64 = SWAR<64, uint64_t>;
 
+static_assert(SWAR<16, u64>::MaxUnsignedLaneValue == 65535);
+static_assert(SWAR<16, u32>::MaxUnsignedLaneValue == 65535);
+static_assert(SWAR<8, u32>::MaxUnsignedLaneValue == 255);
+static_assert(SWAR<4, u32>::MaxUnsignedLaneValue == 15);
+static_assert(SWAR<2, u32>::MaxUnsignedLaneValue == 3);
```
thecppzoo, line 40: Decimal is not the natural base for these tests.
I know we ahve a mess of decimals and hex and binary in these tests, but let's stop the mess.
You may create an assign a PR to me to clean this up if you're upset about this request.

test/swar/BasicOperations.cpp
```diff
@@ -33,6 +33,26 @@ using S32_32 = SWAR<32, uint32_t>;
...
+
+static_assert(SWAR{Literals<8, u32>, {0, 0, 0, 0}}.value() == 0);
+static_assert(SWAR{Literals<8, u32>, {0, 0, 0, 1}}.value() == 1);
```
thecppzoo, line 43: Given the test below, these tests are irrelevant, please eliminate them.

test/swar/BasicOperations.cpp
```diff
@@ -33,6 +33,26 @@ using S32_32 = SWAR<32, uint32_t>;
...
+static_assert(SWAR{Literals<8, u32>, {8, 3, 2, 1}}.value() == 0x08'03'02'01);
+static_assert(SWAR{Literals<8, u32>, {42, 42, 42, 42}}.value() == 0x2A'2A'2A'2A);
+static_assert(SWAR{Literals<4, u32>, {0, 0, 0, 0, 0, 0, 0, 0}}.value() == 0);
+static_assert(SWAR{Literals<4, u32>, {0, 0, 0, 0, 0, 0, 0, 1}}.value() == 1);
+static_assert(SWAR{Literals<4, u32>, {0, 0, 0, 0, 0, 0, 0, 0}}.value() == 0);
```
thecppzoo, line 48: None of these are useful, they are redundant in a bad way

test/swar/BasicOperations.cpp
```diff
@@ -33,6 +33,26 @@ using S32_32 = SWAR<32, uint32_t>;
...
+static_assert(SWAR{Literals<4, u32>, {8, 7, 6, 5, 4, 3, 2, 1}}.value() == 0x8765'4321);
+static_assert(SWAR{Literals<4, u32>, {8, 7, 6, 5, 4, 3, 2, 7}}.value() == 0x8765'4327);
```
thecppzoo, line 50: What do you want to prove or illustrate for the world?

test/swar/BasicOperations.cpp
```diff
@@ -33,6 +33,26 @@ using S32_32 = SWAR<32, uint32_t>;
...
+
+static_assert(BooleanSWAR{Literals<4, u16>, {false, false, false, false}}.value() == 0);
+static_assert(BooleanSWAR{Literals<4, u16>, {true, true, true, true}}.value() == 0b1000'1000'1000'1000);
+static_assert(BooleanSWAR{Literals<8, u32>, {true, true, true, true}}.value() == 0b10000000'10000000'10000000'10000000);
```
thecppzoo, line 54: This is not right, not useful.
1. Make it so the test indicates the endianness
2. Do not miss the opportunity to use different values to illustrate what really is going on.

inc/zoo/swar/SWAR.h
```diff
@@ -231,6 +255,19 @@ template<int NBits, typename T>
...
+        auto result = T{0};
+        for (auto arg : values) {
+            auto bit = arg ? msbOfFirstLane : 0;
+            result = (result << NBits) | bit;
+        }
+        return BooleanSWAR{result};
+    }
+
+    template <std::size_t N, typename = std::enable_if_t<N == Base::Lanes, T>>
+    constexpr BooleanSWAR(Literals_t<NBits, T>, const bool (&values)[N]) : Base(toMsbBools(values)) {}
```
thecppzoo, line 269: Please never follow this antipattern of doing all of the work in the derived class independently of the base class.
Always tie operations in the derived class to the operations in the base class, otherwise, the base class can drift out of alignment.
For example, you may
`Base(Literals<NBits, T>, values) { this->m_v << (NBits - 1); }`
When you have a hierarchical relation between pieces of code, do not fail to make the code in a way in which the relation is fully apparent;; this is the only way to keep the different pieces from eventually violating the relation given enough evolution of the code

thecppzoo (changes requested):

inc/zoo/swar/SWAR.h
jamierpond: Yeah thanks for this feedback, I was also not happy with this for this reason. Will revise.

jamierpond: @thecppzoo see what you think of this?

inc/zoo/swar/SWAR.h
```diff
@@ -68,7 +78,26 @@ struct SWAR {
                 ~(~T(0) << NBits),
         // Use LowerBits in favor of ~MostSignificantBit to not pollute
         // "don't care" bits when non-power-of-two bit lane sizes are supported
-        LowerBits = MostSignificantBit - LeastSignificantBit;
+        LowerBits = MostSignificantBit - LeastSignificantBit,
+        MaxUnsignedLaneValue = LeastSignificantLaneMask;
+
```
jamierpond, line 83: still not sure about the names around `manipulation` etc, but would be good to get your thoughts here.

inc/zoo/swar/SWAR.h
```diff
@@ -68,7 +78,26 @@ struct SWAR {
...
+    template <typename U, typename ManipulationFn>
+    constexpr auto loadBaseTypeIntoLanes(const U (&values)[Lanes],
+                                         const ManipulationFn&& manipulation) {
+        auto result = T{0};
+        for (auto value : values) {
+            auto laneValue = manipulation(value);
+            result = (result << NBits) | laneValue;
+        }
+        return result;
+    }
```
jamierpond, line 93: not 100% sure about the names for `manipulation` etc right now but would be good to get your opinion on this approach.

test/swar/BasicOperations.cpp
jamierpond: I understand why we might want to have just one format, decimal makes the repo a little more approachable and understandable.

inc/zoo/swar/SWAR.h
```diff
@@ -68,7 +78,26 @@ struct SWAR {
                 ~(~T(0) << NBits),
         // Use LowerBits in favor of ~MostSignificantBit to not pollute
         // "don't care" bits when non-power-of-two bit lane sizes are supported
-        LowerBits = MostSignificantBit - LeastSignificantBit;
+        LowerBits = MostSignificantBit - LeastSignificantBit,
+        MaxUnsignedLaneValue = LeastSignificantLaneMask;
+
+    template <typename U, typename ManipulationFn>
+    constexpr auto loadIntoLanes(const U (&values)[Lanes],
+                                 const ManipulationFn&& manipulation) {
```
thecppzoo, line 86: Please never use this indentation style.
One indentation level deeper ought to mean just one level of refinement more.

inc/zoo/swar/SWAR.h
```diff
@@ -68,7 +78,26 @@ struct SWAR {
...
+            auto laneValue = manipulation(value);
+            result = (result << NBits) | laneValue;
+        }
+        return result;
+    }
+
+    template <typename Arg, std::size_t N, typename = std::enable_if_t<N == Lanes, int>>
+    constexpr
+    SWAR(Literals_t<NBits, T>, const Arg (&values)[N]) : m_v{0} {
+        m_v = loadIntoLanes(values, [](auto x) { return x; });
```
thecppzoo, line 98: initialize to the value returned

inc/zoo/swar/SWAR.h
```diff
@@ -231,6 +263,12 @@ template<int NBits, typename T>
 struct BooleanSWAR: SWAR<NBits, T> {
     using Base = SWAR<NBits, T>;
 
+    template <std::size_t N, typename = std::enable_if_t<N == Base::Lanes, T>>
```
thecppzoo, line 266: You may eliminate the errant SFINAE here

inc/zoo/swar/SWAR.h
```diff
@@ -231,6 +263,12 @@ template<int NBits, typename T>
...
+    constexpr BooleanSWAR(Literals_t<NBits, T>, const bool (&values)[N]) : Base{0} {
+        constexpr auto msbOfFirstLane = T{1} << (NBits - 1);
+        this->m_v = Base::loadIntoLanes(values, [](auto x) { return x ? msbOfFirstLane : 0; });
+    }
```
thecppzoo, line 270: This contains the same antipattern.
A `BooleanSWAR` is a SWAR formally of the same base type but whose lanes are a boolean in the most significant bit, the code ought to reflect that.
You did not indicate a response to my comment that this suffices:
`: Base(Literals<NBits, T>, values) { this->m_v << (NBits - 1); }`

test/swar/BasicOperations.cpp
```diff
@@ -5,6 +5,7 @@
 #include <ios>
 #include <iomanip>
 #include <iostream>
+#include <sys/wait.h>
```
thecppzoo, line 8: Why?

test/swar/BasicOperations.cpp
```diff
@@ -33,6 +34,38 @@ using S32_32 = SWAR<32, uint32_t>;
...
+static_assert(SWAR{Literals<4, u8>, {2, 1}}.value() == 0x21);
+static_assert(SWAR{Literals<4, u8>, {1, 2}}.value() == 0x12);
+
+#define F false
+#define T true
+static_assert(BooleanSWAR{Literals<4, u16>, {F, F, F, F}}.value() == 0);
+static_assert(BooleanSWAR{Literals<4, u16>, {T, F, F, F}}.value() == 0b1000'0000'0000'0000);
+static_assert(BooleanSWAR{Literals<4, u16>, {F, T, F, F}}.value() == 0b0000'1000'0000'0000);
+static_assert(BooleanSWAR{Literals<4, u16>, {false, false, true, false}}.value() == 0b0000'0000'1000'0000);
+static_assert(BooleanSWAR{Literals<4, u16>, {false, false, false, true}}.value() == 0b0000'0000'0000'1000);
```
thecppzoo, line 67: Doesn't this set of test look much better than before?

test/swar/BasicOperations.cpp
```diff
@@ -33,6 +34,38 @@ using S32_32 = SWAR<32, uint32_t>;
...
+
```
thecppzoo, line 68: If you #define things for a quick, convenient thing, don't let the macros escape the context where they are useful, practice the hygiene of limiting the lifetime of things to the context where they are useful.  In this case, `#undef`

inc/zoo/swar/SWAR.h
jamierpond: oooh damn, i see... yeah that works

thecppzoo (changes requested):
Improvements all around, like in the movie "Whiplash", not quite my tempo.

Since you have given us the construction from literals, perhaps it is a good idea to do the opposite of having a conversion operator to array of values,
`constexpr operator std::array<T, Lanes>();`
and a named conversion `to_array`

inc/zoo/swar/SWAR.h
```diff
@@ -263,11 +260,9 @@ template<int NBits, typename T>
...
     using Base = SWAR<NBits, T>;
 
-    template <std::size_t N, typename = std::enable_if_t<N == Base::Lanes, T>>
-    constexpr BooleanSWAR(Literals_t<NBits, T>, const bool (&values)[N]) : Base{0} {
-        constexpr auto msbOfFirstLane = T{1} << (NBits - 1);
-        this->m_v = Base::loadIntoLanes(values, [](auto x) { return x ? msbOfFirstLane : 0; });
-    }
+    template <std::size_t N>
+    constexpr BooleanSWAR(Literals_t<NBits, T>, const bool (&values)[N])
+       : Base(Literals<NBits, T>, values) { this->m_v << (NBits - 1); }
```
thecppzoo: My typo in the comment, this has no effect, change to `<<=`

inc/zoo/swar/SWAR.h
jamierpond: Ok I don't understand why this would work? Doesn't seem to produce the correct result. Please feel free to push a change to this branch with something that you think works better.

test/swar/BasicOperations.cpp
jamierpond: Thank you!

test/swar/BasicOperations.cpp
jamierpond: probs an auto import from a typo, rm'd

inc/zoo/swar/SWAR.h
jamierpond: great, works perfectly.

jamierpond:
> Improvements all around, but, like in the movie "Whiplash", still not quite my tempo.
Loving the feedback. Thank you!
> Since you have given us the construction from literals, perhaps it is a good idea to do the opposite of having a conversion operator to array of values, `constexpr operator std::array<T, Lanes>();` and a named conversion `to_array`
I like the idea of this, will check it out.

inc/zoo/swar/SWAR.h
thecppzoo: This might be a reasonable way to do this, but I'd use the appropriate algorithm in `#include <algorithm>` and I would not take by `const`-r-value reference something that ought to be a pure forwarding reference, meaning `ManipulationFN &&`
Make sure you never put `const`-r-value, either `const`-l-value when strictly necessary, or forwarding reference whenever possible.

inc/zoo/swar/SWAR.h
```diff
@@ -81,20 +81,19 @@ struct SWAR {
...
+            result = (result << NBits) | value;
         }
         return result;
     }
 
     template <typename Arg, std::size_t N, typename = std::enable_if_t<N == Lanes, int>>
     constexpr
     SWAR(Literals_t<NBits, T>, const Arg (&values)[N])
-      : m_v{loadIntoLanes(values, [](auto x) { return x; })} {}
+      : m_v{loadIntoLanes(values)} {}
```
thecppzoo: May you please indent to multiples of four spaces?

thecppzoo (approved):
You're free to merge yourself after you're satisfied

inc/zoo/swar/SWAR.h
```diff
@@ -62,6 +63,13 @@ constexpr std::make_unsigned_t<T> lsbIndex(T v) noexcept {
...
 struct SWAR {
     using type = std::make_unsigned_t<T>;
+    constexpr static auto calculateLeastSignificantLaneMask() {
+      const auto m =
+            sizeof(T) * 8 == NBits ? // needed to avoid shifting all bits
+                ~T(0) :
+                ~(~T(0) << NBits);
+      return m;
+    }
     constexpr static inline type
```
thecppzoo, line 73: Nah...
This calculation does not deserve a name.
I think since C++17 we can use constexpr lambdas, that would be more appropriate below.
However, I showed you that this calculation is not needed, you're repating, and that's the bad part, the bulk of the work done in other constants.
Better than repeating the work is to use the work already done, primarily because then you have cross-checking that things work, if there's a mistake in the old, the new might discover it, if further nuance is discovered the new might benefit from it directly.

thecppzoo (changes requested):
The latest changes made the PR not mergeable anymore, please review

inc/zoo/swar/SWAR.h
```diff
@@ -94,6 +99,15 @@ struct SWAR {
...
 
+    constexpr std::array<T, Lanes> to_array() const noexcept {
+        std::array<T, Lanes> result;
+        for (int i = 0; i < Lanes; ++i) {
+            auto otherEnd = Lanes - i - 1;
+            result[otherEnd] = at(i);
+        }
+        return result;
+    }
+
```
thecppzoo: This endianness conversion bothers me.
Your code is correct, but there is something about endianness that is creeping up that is bothering me.

test/CMakeLists.txt
```diff
@@ -81,7 +81,7 @@ if(MSVC)
     endif()
 else()
     # Non-MSVC specific configuration (original content)
-    set(CMAKE_CXX_STANDARD 17)
+    set(CMAKE_CXX_STANDARD 23)
```
thecppzoo, line 84: Support for compilation under MSVC is on a "best effort" basis, we try, but we can't guarantee we will succeed at all the things.  It is acceptable to bump to 23, but please explain why you found the need

jamierpond: This is for the non-MSVC path. This is to enable support for std::array in a constexpr context.

inc/zoo/swar/SWAR.h
jamierpond: Thanks, should have reverted. This was a debugging attempt. Was getting a red herring compiler message about this variable not being constexpr when trying to implement `to_array`.

inc/zoo/swar/SWAR.h
```diff
@@ -68,7 +78,21 @@ struct SWAR {
                 ~(~T(0) << NBits),
         // Use LowerBits in favor of ~MostSignificantBit to not pollute
         // "don't care" bits when non-power-of-two bit lane sizes are supported
-        LowerBits = MostSignificantBit - LeastSignificantBit;
+        LowerBits = MostSignificantBit - LeastSignificantBit,
+        MaxUnsignedLaneValue = LeastSignificantLaneMask;
+
+    template <typename U>
+    constexpr auto loadIntoLanes(const U (&values)[Lanes]) const noexcept {
```
thecppzoo, line 85: Please rename to `from_array`.  As the symmetrical operation to `to_array`.
Also, this is a factory, it ought to be `static`, a class-function, not instance.

inc/zoo/swar/SWAR.h
thecppzoo: Your code is correct.  Furthermore, your choice to adopt big endianness for arrays was superior to the alternative, but now we have this flipping of endianness that I don't know what to do about; I'm just voicing the idea, and plan to keep it in the back of my head to see where we go with this.

thecppzoo:
My preference with this PR is to not just introduce the support for direct initialization of literals, but complete it with the following:
1. Making the essential function `from_array` a part of the public interface (we ought to indicate the semantics is that of big endianness in the comments inline with the code), as a class-function, because it is a factory.
2. Keep the `Literals` infrastructure you already made, extend with a constructor from `std::array`
3. `to_array`

Separately, I think we can also add a header, such as `SWAR_IO.h` where the insertion and extraction operators are implemented.  I don't think these things are unrelated: The ergonomics of the literals are highly cohesive with stream insertion and extraction.  We need this "yesterday" for debugging and tracing.

Indicate if you'd like to split these features with me or prefer to do them all yourself.

jamierpond:
If you want to take on the `std::array` stuff and getting that to that could be nice. 

From what I understood from playing around with this today we will require C++ >17 in order to use `constexpr std::array` in `static_assert` tests, but this may be a limitation of my understanding. 

Even with C++23 am seeing failures to compiler on `g++`.

inc/zoo/swar/SWAR.h
```diff
@@ -68,7 +76,21 @@ struct SWAR {
...
         // "don't care" bits when non-power-of-two bit lane sizes are supported
-        LowerBits = MostSignificantBit - LeastSignificantBit;
+        LowerBits = MostSignificantBit - LeastSignificantBit,
+        MaxUnsignedLaneValue = LeastSignificantLaneMask;
+
+    template <typename U>
+    constexpr static auto from_array(const U (&values)[Lanes]) noexcept {
+        auto result = T{0};
+        for (auto value : values) {
+            result = (result << NBits) | value;
```
Scottbruceheart, line 86: blitElement?

inc/zoo/swar/SWAR.h
```diff
@@ -12,6 +12,14 @@
 
 namespace zoo { namespace swar {
 
+template <int NBits, typename T>
+struct SWAR;
+
+template <int NumBits, typename BaseType> struct Literals_t {};
+
+template <int NumBits, typename BaseType>
+constexpr Literals_t<NumBits, BaseType> Literals{};
```
Scottbruceheart, line 21: I dislike that we've got to do a packaged sub-type to...get a type match? I haven't tried to avoid that before, but it seems like something probably tractable.

inc/zoo/swar/SWAR.h
thecppzoo: I this does not quite blit the element, if it will move in an element, it shifts up what it got and then adds to the least significant lane

thecppzoo (approved):

PR #81, opened 2024-04-19
Add `SWAR::to_array` to complement literals work.

jamierpond:


thecppzoo:
To **complEment**, not "compliment"

test/CMakeLists.txt
```diff
@@ -81,7 +81,7 @@ if(MSVC)
     endif()
 else()
     # Non-MSVC specific configuration (original content)
-    set(CMAKE_CXX_STANDARD 17)
+    set(CMAKE_CXX_STANDARD 23)
```
thecppzoo, line 84: This is not right.

jamierpond: This is what I mentioned on the other PR, I think we will need to use at least C++ 20 to gain access to `constexpr` `std::array`. Maybe you want to have a look at the `std::array` component of this? I'll be signing off right now since is very late here in the UK!

inc/zoo/swar/SWAR.h
```diff
@@ -73,10 +73,14 @@ struct SWAR {
         AllOnes = ~std::make_unsigned_t<T>{0} >> PaddingBitsCount, // Also constructed in RobinHood utils: possible bug?
         LeastSignificantBit = meta::BitmaskMaker<T, std::make_unsigned_t<T>{1}, NBits>::value,
         MostSignificantBit = LeastSignificantBit << (NBits - 1),
-        LeastSignificantLaneMask =
-            sizeof(T) * 8 == NBits ? // needed to avoid shifting all bits
-                ~T(0) :
-                ~(~T(0) << NBits),
+        LeastSignificantLaneMask = []() {
+          constexpr auto NBitsLessThanByteOfT = NBits < sizeof(T) * 8;
```
thecppzoo: `NBitsLessThanBitsOfT`

inc/zoo/swar/SWAR.h
```diff
@@ -73,10 +73,14 @@ struct SWAR {
...
+          if constexpr (NBitsLessThanByteOfT) {
+            return (T(1) << NBits) - 1;
+          } else {
+            return ~T(0);
+          }
+        }(),
```
thecppzoo: Indentation?

inc/zoo/swar/SWAR.h
jamierpond: rm'd name

test/swar/BasicOperations.cpp
```diff
@@ -54,7 +54,22 @@ static_assert(SWAR{Literals<4, u8>, {1, 2}}.value() == 0x12);
...
-static_assert(SWAR{Literals<4, u8>, {1, 2}}.to_array() == std::array<u8, 2>{1, 2});
+template <size_t N, typename A, typename B>
+constexpr auto compareContents(A a, B b) {
+    for (auto i = 0; i < N; ++i) {
+        if (a[i] != b[i]) {
+            return false;
+        }
+    }
+    return true;
+}
```
thecppzoo: `compareContents` of what? what's the comparison?
I know what you're trying to solv,e the lack of `constexpr` operator `==` for `std::array`.
Either solve that specifically, or, if better, you try to generalize, then really generalize.  I think your aim was toward equality, then by all means, implement the equality operator for things that resemble an `std::array`. 
This utility function is very needed.
`std::equal` almost does what is needed, but it has two design mistakes:
1. It does not check the size of the ranges
2. It wasn't `constexpr` from the beginning.
You may write `operator==`  for `T (&a)[Length]` and `std::array<T, Length>` using an implementation that is common to both.

This repository has a problem with half-thought code.

You don't need C++ 20 anymore with this utility, right?

inc/zoo/swar/SWAR.h
```diff
@@ -73,10 +73,14 @@ struct SWAR {
         AllOnes = ~std::make_unsigned_t<T>{0} >> PaddingBitsCount, // Also constructed in RobinHood utils: possible bug?
         LeastSignificantBit = meta::BitmaskMaker<T, std::make_unsigned_t<T>{1}, NBits>::value,
         MostSignificantBit = LeastSignificantBit << (NBits - 1),
-        LeastSignificantLaneMask =
-            sizeof(T) * 8 == NBits ? // needed to avoid shifting all bits
-                ~T(0) :
-                ~(~T(0) << NBits),
+        LeastSignificantLaneMask = []() {
+          constexpr auto NBitsLessThanByteOfT = NBits < sizeof(T) * 8;
+          if constexpr (NBitsLessThanByteOfT) {
```
thecppzoo: Why using `if constexpr`  while you could use the ternary operator, `?`?

test/swar/BasicOperations.cpp
```diff
@@ -68,7 +68,7 @@ using S4U8 = SWAR<4, u8>;
 constexpr auto A = S4U8{Literals<4, u8>, {4, 4}};
 constexpr auto B = std::array<u8, 2>{4, 4};
 
-static_assert(compareContents<S4U8::Lanes, std::array<u8, 2>>(A.to_array(), B));
+static_assert(compareContents<S4U8::Lanes>(A.to_array(), B));
```
thecppzoo: I don't think giving the names `A` and `B` help clarity in any way.
It seems to me the defining expressions for `A` and `B` are clearer than splitting them into a `A` and `B`.

inc/zoo/swar/SWAR.h
```diff
@@ -85,15 +85,24 @@ struct SWAR {
...
+        for (; first != last; ++first) {
+            result = (result << NBits) | *first;
         }
-        return SWAR{result};
+        return result;
     }
 
+    template <typename U>
+    constexpr static auto from_array(const U (&values)[Lanes]) noexcept {
+        return SWAR{from_range(std::begin(values), std::end(values))};
```
thecppzoo, line 99: NOOOOOO!!!!
do not qualify `begin` and `end` with `std::`!!!!!
We define `begin` and `end` within zoo for zoo's own containers.
The same with `swap`.
They are library customization points.
The correct way is to do this, whenever possible, it is fine if you don't do this whenever not possible:
```c++
constexpr static auto from_array(const U (&values)[Lanes]) noexcept {
    using std::begin; using std::end; // fallback to std
    yada yada begin(range), yada yada end(range)
```

inc/zoo/swar/SWAR.h
```diff
@@ -85,15 +85,24 @@ struct SWAR {
...
+    }
+
+    using ArrayType = std::array<T, Lanes>;
+
+    constexpr SWAR(const ArrayType& array) : m_v{from_range(array.begin(), array.end())} {}
```
thecppzoo, line 104: You're again using L-value references gratuitously.
Let's say you're passed a temporary `std::vector` , since you explicitly request the L-value reference here, a totally unnecessary copy of the vector will be made, and the compiler won't save you from the folly since an allocation is potentially involved, the compiler won't be able to prove no allocation will be called and then it will have to preserve the potential side effects of the allocation.

thecppzoo: Please put the identifier modifiers next to the identifier they modify.
`ArrayType &array`
I don't know why, given the rules of the language, the style of putting the modifiers to the type is overwhelmingly more popular, since they don't bind to the type, but to the identifier being declared, they simply don't apply to the type!

inc/zoo/swar/SWAR.h
jamierpond: this apparently was required to appease the compiler.

jamierpond:
@thecppzoo could you resolve comments which you think are resolved please?

test/CMakeLists.txt
thecppzoo: I don't understand what you mean, `std::array` has been `constexpr` from the start in '11

inc/zoo/swar/SWAR.h
thecppzoo: Indentation?

inc/zoo/swar/SWAR.h
```diff
@@ -62,18 +73,82 @@ struct SWAR {
...
+                return ~T(0);
+            }
+        }(),
         // Use LowerBits in favor of ~MostSignificantBit to not pollute
         // "don't care" bits when non-power-of-two bit lane sizes are supported
-        LowerBits = MostSignificantBit - LeastSignificantBit;
+        LowerBits = MostSignificantBit - LeastSignificantBit,
+        MaxUnsignedLaneValue = LeastSignificantLaneMask;
+
+    template <typename InputIt>
```
thecppzoo, line 88: Do you mean this is for any input iterator?
Do we have a unit test that uses input interators?
Also, unfortunately, `from_range` is not an appropriate identifier, since we are using iterator pairs, I believe that the convention emerged to name things that take a single-argument range as the suffix `range`

inc/zoo/swar/SWAR.h
```diff
@@ -62,18 +73,82 @@ struct SWAR {
...
+    constexpr static auto from_array(const U (&values)[Lanes]) noexcept {
+        using std::begin; using std::end;
+        return SWAR{from_range(begin(values), end(values))};
+    }
+
+    template <typename U>
+    constexpr static auto from_array(const std::array<T, Lanes> &values) noexcept {
+        using std::begin; using std::end;
+        return SWAR{from_range(begin(values), end(values))};
+    }
```
thecppzoo, line 107: I believe we can have a `from_range(Range` overload and not make it specific to `std::array` unnecessarily

inc/zoo/swar/SWAR.h
```diff
@@ -62,18 +73,82 @@ struct SWAR {
...
+    constexpr
+    SWAR(Literals_t<NBits, T>, const Arg (&values)[N]) : m_v{from_array(values)} {}
+
+    constexpr std::array<T, Lanes> to_array() const noexcept {
+        std::array<T, Lanes> result = {};
+        for (int i = 0; i < Lanes; ++i) {
+            auto otherEnd = Lanes - i - 1;
+            result[otherEnd] = at(i);
+        }
+        return result;
```
thecppzoo, line 123: I think this can be encapsulated into an expression

inc/zoo/swar/SWAR.h
```diff
@@ -62,18 +73,82 @@ struct SWAR {
...
+//     constexpr auto operator==(T (&values)[Lanes]) const noexcept {
+//         return compareToContainer(values);
+//     }
+//
+//     constexpr auto operator==(std::array<T, Lanes> values) const noexcept {
+//         return compareToContainer(values);
+//     }
+
+    template <typename B>
+    constexpr bool compareToContainer(B b) const noexcept {
```
thecppzoo, line 139: what comparison?

inc/zoo/swar/SWAR.h
```diff
@@ -62,18 +73,82 @@ struct SWAR {
...
+        auto a = to_array();
```
thecppzoo, line 140: No, this is not acceptable.

inc/zoo/swar/SWAR.h
```diff
@@ -305,6 +390,9 @@ struct BooleanSWAR: SWAR<NBits, T> {
     convertToBooleanSWAR(SWAR<NB, TT> arg) noexcept;
 };
 
+template <int NBits, typename T>
+BooleanSWAR(Literals_t<NBits, T>, const bool (&values)[BooleanSWAR<NBits, T>::Lanes]) -> BooleanSWAR<NBits, T>;
```
thecppzoo, line 394: Let's respect 80 columns

test/swar/BasicOperations.cpp
```diff
@@ -81,66 +162,62 @@ HE(2, u8, 0xAA, 0x2);
...
-        multiplication_OverflowUnsafe_SpecificBitCount_deprecated<4>(
-            Micand, Mplier
-        );
-    CHECK(Expected == result.value());
+  SWAR<8, u32> Micand{0x5030201};
+  SWAR<8, u32> Mplier{0xA050301};
+  auto Expected = 0x320F0601;
+  auto result = multiplication_OverflowUnsafe_SpecificBitCount_deprecated<4>(
+      Micand, Mplier);
+  CHECK(Expected == result.value());
```
thecppzoo, line 170: Indentation!

test/swar/BasicOperations.cpp
```diff
@@ -81,66 +162,62 @@ HE(2, u8, 0xAA, 0x2);
...
-        u64
-            input =   0b1010'1001'0110'0001'1001'0000'0010'1010'0100'0111'1110'1001'1111'0001'1110'1011,
-            mask =    0b0110'0000'0001'0101'0101'1111'0101'1100'0110'1111'0100'0111'0001'1000'0101'0010,
-            expected =0b0001'0000'0000'0001'0001'0000'0000'0010'0010'0111'0001'0001'0001'0000'0010'0001;
-        using S = S4_64;
-        auto v = compress(S{input}, S{mask});
-        CHECK(expected == v.value());
-    }
+  unsigned Mask = 0b0001'0011'0111'0111'0110'1110'1100'1010,
+           ToMove = 0b0101'0101'0101'0101'0101'0101'0101'0101,
```
thecppzoo, line 205: No!!!
That accidental alignment style is not acceptable!!

test/swar/BasicOperations.cpp
```diff
@@ -258,200 +335,173 @@ static_assert(0x0808'0808 == u32(broadcast<8>(SWAR<8, u32>(0x0000'0008))));
...
-static_assert(0x00000008 == greaterEqual<7>(SWAR<4, uint32_t>(0x0123'4567)).value());
-static_assert(0x80000000 == greaterEqual<7>(SWAR<4, uint32_t>(0x7654'3210)).value());
+static_assert(0x80880008 == greaterEqual<3>(SWAR<4,
+uint32_t>(0x3245'1027)).value()); static_assert(0x88888888 ==
+greaterEqual<0>(SWAR<4, uint32_t>(0x0123'4567)).value());
+static_assert(0x88888888 == greaterEqual<0>(SWAR<4,
+uint32_t>(0x7654'3210)).value()); static_assert(0x00000008 ==
+greaterEqual<7>(SWAR<4, uint32_t>(0x0123'4567)).value());
+static_assert(0x80000000 == greaterEqual<7>(SWAR<4,
+uint32_t>(0x7654'3210)).value());
```
thecppzoo, line 357: what the fuck?

inc/zoo/swar/SWAR.h
jamierpond: rm'd this, since now unused, was part of a draft

inc/zoo/swar/SWAR.h
jamierpond: rm'd

test/swar/BasicOperations.cpp
```diff
@@ -258,200 +335,173 @@ static_assert(0x0808'0808 == u32(broadcast<8>(SWAR<8, u32>(0x0000'0008))));
...
-    0x0123'4567,
-    0x8888'8888)
+GE_MSB_TEST(0x1000'0010, 0x0111'1101, 0x8000'0080)
+GE_MSB_TEST(0x4333'3343, 0x4444'4444, 0x8000'0080)
+GE_MSB_TEST(0x0550'0110, 0x0110'0550, 0x8888'8008)
+GE_MSB_TEST(0x4771'1414, 0x4641'1774, 0x8888'8008)
+GE_MSB_TEST(0x0123'4567, 0x0000'0000, 0x8888'8888)
+GE_MSB_TEST(0x0123'4567, 0x7777'7777, 0x0000'0008)
+GE_MSB_TEST(0x0000'0000, 0x0123'4567, 0x8000'0000)
+GE_MSB_TEST(0x7777'7777, 0x0123'4567, 0x8888'8888)
```
thecppzoo, line 372: Do you think Scott did the multi-line gratuitously?

test/swar/BasicOperations.cpp
jamierpond: gah sorry this is my auto-formatter. i think it would be valuable to have a clang-format to avoid style issues like this in the future. it at least is valuable to have this to minimise the frequency of merge conflicts.

test/swar/BasicOperations.cpp
jamierpond: auto-formatter, see above.

test/CMakeLists.txt
jamierpond: this change is outdated.

inc/zoo/swar/SWAR.h
jamierpond: you think this should be in some named expression like `convertSwarToArray` ?

inc/zoo/swar/SWAR.h
jamierpond: Could you propose an alternative?

inc/zoo/swar/SWAR.h
jamierpond: Yes, I think that would be good. Btw I didn't find the SWAR `begin` and `end` functions I think you mentioned in another comment. Was wondering if you could explain more about that?

test/swar/BasicOperations.cpp
jamierpond: thanks for pointing out the formatting thing, will try and disable my editor from auto-formatting

test/swar/BasicOperations.cpp
thecppzoo: The question I asked was to help not missing the point: whitespace is a communication channel.  Scott put this in lines so that one could easily see the three things, the two values under comparison and the result.
The problem with using auto formatting is that it destroys whitespace as a communications channel.  That's the problem.

inc/zoo/swar/SWAR.h
```diff
@@ -12,6 +13,16 @@
 
 namespace zoo { namespace swar {
 
+template <int NBits, typename T>
+struct SWAR;
+
+template <int NumBits, typename BaseType> struct Literals_t {
+  constexpr static void (SWAR<NumBits, BaseType>::*value)() = nullptr;
```
thecppzoo, line 20: indentation

inc/zoo/swar/SWAR.h
```diff
@@ -62,13 +73,53 @@ struct SWAR {
...
-        LeastSignificantLaneMask =
-            sizeof(T) * 8 == NBits ? // needed to avoid shifting all bits
-                ~T(0) :
-                ~(~T(0) << NBits),
+        LeastSignificantLaneMask = []() {
+            if constexpr (NBits < sizeof(T) * 8) {
+                return (T(1) << NBits) - 1;
+            } else {
+                return ~T(0);
+            }
```
thecppzoo, line 81: What is the reason for this `if constexpr`? because the ternary operator `?` would get the compiler to warn on the branch that would not be active?

inc/zoo/swar/SWAR.h
thecppzoo: After I wrote the comment I thought about it and thought that we can keep `from_range`, these functions that take a pair of iterators ought to have been named ranges from the start, then it is legitimate to retrofit the particle "range" into identifiers.

inc/zoo/swar/SWAR.h
jamierpond: At one point the compiler was failing at the ternary operator, so I moved it over to this.

jamierpond:
@thecppzoo is this good to merge?

inc/zoo/swar/SWAR.h
thecppzoo: Yes, there is no option.

inc/zoo/swar/SWAR.h
```diff
@@ -367,7 +367,10 @@ struct BooleanSWAR: SWAR<NBits, T> {
 };
 
 template <int NBits, typename T>
-BooleanSWAR(Literals_t<NBits, T>, const bool (&values)[BooleanSWAR<NBits, T>::Lanes]) -> BooleanSWAR<NBits, T>;
+BooleanSWAR(
+    Literals_t<NBits, T>,
+    const bool (&values)[BooleanSWAR<NBits, T>::Lanes])
+  -> BooleanSWAR<NBits, T>;
```
thecppzoo, line 374: Respecting 80 columns is of less importance than having a consistent way to express "detail".
Closing the list of arguments in this deduction guide far to the right breaks the context of "where does the list of arguments close?":
```
BooleanSWAR(
    Literals_t<NBits, T>,
    const bool (&values)[BooleanSWAR<NBits, T>::Lanes]
)
```
I don't understand your insistence in doing things in a different way.
This then goes to worsen the layout by introducing a two-space accidental indentation level.
Again: please desist.

thecppzoo (changes requested):

inc/zoo/swar/SWAR.h
thecppzoo: I did not understadn your question, this is not about a name or identifier, but something like
```
template<yada yada>
some_converter_name_impl(std::index_sequence<Indices...>, arguments) {
    return { at(Indices)...};
}
```
where `Indices` is an argument pack of numbers 0 to the lane count

thecppzoo:
We are almost there.
Take into account the outstanding comments.
After merge, I will use the expression form using an `std::index_sequence`, then whoever merges this PR should create an issue for me to do that.

inc/zoo/swar/SWAR.h
jamierpond: Not really following what you're getting at here? Would this be for going to or from an array?  
We could substitude doing the weird array packs we're been doing and use `integer_sequence` instead, possibly?

inc/zoo/swar/SWAR.h
jamierpond: No dramas, will update. For your sanity I've updated my indentation level for this project specifically.

thecppzoo: You have said something similar three times.
Desist?
What about giving a chance to expressing something with whitespace? could that help you see that clang-format is made by people who don't see whitespace as a communication channel, for more people of the same?
What about learning about "tools that help you make the wrong thing easier", Brit Kevlin Henney?

inc/zoo/swar/SWAR.h
```diff
@@ -62,13 +74,53 @@ struct SWAR {
...
+                return ~T(0);
+            }
+        }(),
         // Use LowerBits in favor of ~MostSignificantBit to not pollute
         // "don't care" bits when non-power-of-two bit lane sizes are supported
-        LowerBits = MostSignificantBit - LeastSignificantBit;
+        LowerBits = MostSignificantBit - LeastSignificantBit,
+        MaxUnsignedLaneValue = LeastSignificantLaneMask;
+
+    template <typename InputIt>
```
thecppzoo, line 89: Resolved the other comment, the aspect remaining is whether input iterators is meant here.
If you want to mean "iterator to the input", what is meant is the input iterator category
https://en.cppreference.com/w/cpp/named_req/InputIterator
This code can indeed work with input iterators.  I'm asking if you noticed that `InputIt` has a particular meaning.

inc/zoo/swar/SWAR.h
thecppzoo: Not in SWAR, but they could be implemented... the thing is what ought to be the type of a SWAR iterators to iterate over all the lanes? I think iterating serially over the lanes is of low priority.
This overload should be turned into this:
```suggestion
    template <typename Range>
    constexpr static auto from(const Range &values) noexcept {
        using std::begin; using std::end;
        return SWAR{from_range(begin(values), end(values))};
    }
```

inc/zoo/swar/SWAR.h
```diff
@@ -62,13 +74,53 @@ struct SWAR {
...
+    constexpr static auto from_range(InputIt first, InputIt last) noexcept {
```
thecppzoo, line 90: I forgot to ask much earlier why `snake_case` instead of `camelCase`

thecppzoo, line 90: is `from_range` a developing consensus concerning constructors from ranges?
I think `range` is not necessary.
Also, I am suspicious of these pseudo-constructors (you need to call them from the type as in `SWAR<yada yada>::from_range)` these are like something we can call "named constructors".  Do you have inspiration from some other code about this idiom?

inc/zoo/swar/SWAR.h
thecppzoo: Leave it as it is, if you want.
I will code from memory, something like this:
namespace impl {
template<typename T, std::size_t... Indices>
constexpr
std::array<T, sizeof...(Indices)> to_array(SWAR<?????> s, std::index_sequence<Indices...>) noexcept {
    return { s.at(Indices)... };
}
}
// ...
constexpr std::array<T, Lanes> to_array() const noexcept {
    return impl::to_array(*this, std::make_index_sequence<Lanes>{});
}
```
I've always thought of this idiom, that to give it a name I called it "expression style" as more elegant; I love translating the minutiae of iteration to compile-time-only code.
I also dislike the imperative style.

File: inc/zoo/swar/SWAR.h (hunk 10 above)
Comment (jamierpond) [reply inc/zoo/swar/SWAR.h:108]: I seee, nice.

File: inc/zoo/swar/SWAR.h [hunk 24]
```diff
@@ -161,6 +213,12 @@ struct SWAR {
     T m_v;
 };
 
+template <int NBits, typename T, typename Arg>
+SWAR(Literals_t<NBits, T>, const Arg (&values)[SWAR<NBits, T>::Lanes]) -> SWAR<NBits, T>;
```
thecppzoo, line 217:
```suggestion
SWAR(Literals_t<NBits, T>, const Arg (&values)[SWAR<NBits, T>::Lanes])
-> SWAR<NBits, T>;
```

thecppzoo:
```suggestion
SWAR(
    Literals_t<NBits, T>, const Arg (&values)[SWAR<NBits, T>::Lanes]
) -> SWAR<NBits, T>;
```

thecppzoo:
```suggestion
SWAR(
    Literals_t<NBits, T>,
    const Arg (&values)[SWAR<NBits, T>::Lanes]
) -> SWAR<NBits, T>;
```

thecppzoo: You can also create a new line and indentation level each `<` of each template.
I would not mind any of those.
And you can use K&R, as in
```
void function()
{
    // ...
} 
```

inc/zoo/swar/SWAR.h
jamierpond: Let's do this in another PR.
I think Scott would like to start using the ergonomics in writing some tests. 
It will be valueable to let Scott's magnetic field to push back on this code's electrical field.

inc/zoo/swar/SWAR.h
jamierpond: I feel like I've seen a few patterns like this, especially in unit conversion. 
I think it reads nicely. 

It could be valuable for you to create a supplementary PR for this to: 
- make this more real constructor centric
- do the compile-time index_sequence impl

what do you think ?

inc/zoo/swar/SWAR.h
jamierpond: i didn't notice. did you have another idea in mind?

test/swar/BasicOperations.cpp
```diff
@@ -33,6 +31,71 @@ using S32_32 = SWAR<32, uint32_t>;
...
+static_assert(SWAR{Literals<16, u64>, {1, 2, 3, 4}}.at(1) == 3);
+
+// Macro required because initializer lists are not constexpr
+#define ARRAY_TEST(SwarType, ...)                                              \
+    static_assert([]() {                                                       \
+        using S = SwarType;                                                    \
+        constexpr auto arry = std::array{__VA_ARGS__};                         \
+        constexpr auto test_array = S{S::Literal, {__VA_ARGS__}}.to_array();   \
+        static_assert(arry.size() == S::Lanes);                                \
+        for (auto i = 0; i < S::Lanes; ++i) {                                  \
```
Scottbruceheart, line 70: does operator=() not work correctly for std::array?

test/swar/BasicOperations.cpp
```diff
@@ -33,6 +31,71 @@ using S32_32 = SWAR<32, uint32_t>;
...
 using S64_64 = SWAR<64, uint64_t>;
 
+static_assert(SWAR<16, u64>::MaxUnsignedLaneValue == 65535);
+static_assert(SWAR<16, u32>::MaxUnsignedLaneValue == 65535);
+static_assert(SWAR<8, u32>::MaxUnsignedLaneValue == 255);
+static_assert(SWAR<4, u32>::MaxUnsignedLaneValue == 15);
+static_assert(SWAR<2, u32>::MaxUnsignedLaneValue == 3);
+
+static_assert(SWAR{Literals<32, u64>, {2, 1}}.value() == 0x00000002'00000001);
+static_assert(SWAR{Literals<32, u64>, {1, 2}}.value() == 0x00000001'00000002);
```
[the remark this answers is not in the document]
jamierpond: greatttt point to be fair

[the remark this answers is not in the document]
jamierpond: done

test/swar/BasicOperations.cpp
jamierpond: it would seem not for constexpr things in cpp17

jamierpond:
@Scottbruceheart once again, I believe this is good to go.

inc/zoo/swar/SWAR.h
Scottbruceheart: Is this comment addressed? I think Eduardo's response is up.

PR #83, opened 2024-05-22
Implementation of atol using SWAR of 128 bits

thecppzoo:
This is the first time we apply `SWAR` to a base type of 128 bits.  There are many inefficiencies about using `__uint128_t` for the base type of `SWAR`s, since that form of 128 bit integers is emulated by the compiler using two 64 bit values.

This PR concerns itself with bare-minimum support of 128 bits, which only required modifying the operation of counting trailing zeros, to apply it to our string benchmarks to have a performance idea.

The changes introduced are worthy of being upgraded to "master", but they are not consistent with full support of `SWAR` of 128 bit base.

### DIFF
--- a/benchmark/atoi-catch2-tests.cpp
+++ b/benchmark/atoi-catch2-tests.cpp
@@ -3,9 +3,19 @@
 #include "zoo/swar/SWAR.h"
 
 uint64_t calculateBase10(zoo::swar::SWAR<8, __uint128_t>) noexcept;
+namespace zoo {
+int64_t c_strToL(const char *str) noexcept;
+}
+
+template<int RequiredAlignment>
+void requireAlignment(const void *p) {
+    std::uintptr_t asNumber = reinterpret_cast<std::uintptr_t>(p);
+    REQUIRE(asNumber % RequiredAlignment == 0);
+}
 
 TEST_CASE("Calculate Base10", "[pure-test][swar][atoi]") {
-    alignas(16) auto inputString = "1234567898765432";
+    alignas(16) char inputString[] = "1234567898765432";
+    requireAlignment<16>(inputString);
     using S = zoo::swar::SWAR<8, __uint128_t>;
     S input;
     memcpy(&input.m_v, inputString, 16);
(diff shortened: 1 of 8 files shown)


benchmark/atoi.cpp
```diff
@@ -53,6 +54,19 @@ uint32_t calculateBase10(zoo::swar::SWAR<8, uint64_t> convertedToIntegers) noexc
     return uint32_t(by10001base2to32.value() >> 32);
 }
 
+uint64_t calculateBase10(zoo::swar::SWAR<8, __uint128_t> convertedToIntegers) noexcept {
+    auto by11base256 = convertedToIntegers.multiply(256*10 + 1);
+    auto bytePairs = zoo::swar::doublePrecision(by11base256).odd;
+    //static_assert(std::is_same_v<decltype(bytePairs), zoo::swar::SWAR<16, uint64_t>>);
```
Scottbruceheart, line 60: Is this meant to be submitted?

benchmark/atoi_impl.h
```diff
@@ -0,0 +1,35 @@
...
+
+namespace zoo {
+
+/// @brief Loads the "block" containing the pointer, by proper alignment
+/// @tparam PtrT Pointer type for loading
+/// @tparam Block as the name indicates
+/// @param pointerInsideBlock the potentially misaligned pointer
+/// @param b where the loaded bytes will be put
+/// @return a pair to indicate the aligned pointer to the base of the block
+/// and the misalignment, in bytes, of the source pointer
```
Scottbruceheart, line 22: For absolute clarity, specify the range misalignment can take. It probably is not [-maxint, +maxint]

benchmark/atoi.cpp
```diff
@@ -53,6 +54,19 @@ uint32_t calculateBase10(zoo::swar::SWAR<8, uint64_t> convertedToIntegers) noexc
...
+    auto by101base2to16 = bytePairs.multiply(1 + (100 << 16));
+    auto byteQuads = zoo::swar::doublePrecision(by101base2to16).odd;
+    auto by10001base2to32 = byteQuads.multiply(1 + (10000ull << 32));
+    auto byteOcts = zoo::swar::doublePrecision(by10001base2to32).odd;
+    auto byHundredMillionBase2to64 =
```
Scottbruceheart, line 65: A comment along the lines of 'extends the 8 byte technique from 10000s to 100,000,000s' seems useful.

benchmark/atoi-catch2-tests.cpp
```diff
@@ -0,0 +1,18 @@
+#include "catch2/catch.hpp"
+
+#include "zoo/swar/SWAR.h"
+
+uint64_t calculateBase10(zoo::swar::SWAR<8, __uint128_t>) noexcept;
+
+TEST_CASE("Calculate Base10", "[pure-test][swar][atoi]") {
```
Scottbruceheart, line 7: Perhaps more tests? something unaligned?

benchmark/atoi.cpp
```diff
@@ -173,11 +168,33 @@ auto leadingDigitsCount(const char *p) noexcept {
...
+    using Traits = ConversionTraits<Result>;
+    typename Traits::PowersOf10Array rv{1};
+    for (std::size_t i = 1; i < Traits::NPositions; ++i) {
+        rv[i] = rv[i - 1] * 10;
+    }
+    return rv;
+};
+
+template<typename Return>
+Return c_strToIntegral(const char *str) noexcept {
```
Scottbruceheart, line 196: A test for the version that handles 16 byte chunks that crosses a 16 byte boundary seems wise, but perhaps I missed it.

benchmark/atoi.cpp
thecppzoo: No, thanks.

benchmark/atoi-catch2-tests.cpp
thecppzoo: Yes, the `atol` that straddles two blocks was unit-tested

PR #86, opened 2024-05-26
Make power of two check 

jamierpond:
- **new literals**
- **PR feedback**
- **tidy tests**
- **tidy reused code for booleanswaer**
- **update name**
- **style**
- **test style**
- **fix**
- **undef util**
- **simplify again**
- **fmt**
- **Add to Array**
- **rename**
- **nailed it**
- **cleanup**
- **array tests?**
- **indentaion**
- **update indent**
- **from array**
- **more updates**
- **wip**
- **rm unused**
- **respect 80 chars**
- **undo clang formatting**
- **format boolean swar**
- **improvementws**
- **indentation**
- **make sure we understand equality**
- **works**

### DIFF
--- a/inc/zoo/swar/SWAR.h
+++ b/inc/zoo/swar/SWAR.h
@@ -82,8 +82,8 @@ struct SWAR {
         MaxUnsignedLaneValue = LeastSignificantLaneMask;
 
     template <typename U, typename ManipulationFn>
-    constexpr auto loadBaseTypeIntoLanes(const U (&values)[Lanes],
-                                         const ManipulationFn&& manipulation) {
+    constexpr auto loadIntoLanes(const U (&values)[Lanes],
+                                 const ManipulationFn&& manipulation) {
         auto result = T{0};
         for (auto value : values) {
             auto laneValue = manipulation(value);
@@ -95,7 +95,7 @@ struct SWAR {
     template <typename Arg, std::size_t N, typename = std::enable_if_t<N == Lanes, int>>
     constexpr
     SWAR(Literals_t<NBits, T>, const Arg (&values)[N]) : m_v{0} {
-        m_v = loadBaseTypeIntoLanes(values, [](auto x) { return x; });
+        m_v = loadIntoLanes(values, [](auto x) { return x; });
     }
 
 
@@ -266,7 +266,7 @@ struct BooleanSWAR: SWAR<NBits, T> {
     template <std::size_t N, typename = std::enable_if_t<N == Base::Lanes, T>>
     constexpr BooleanSWAR(Literals_t<NBits, T>, const bool (&values)[N]) : Base{0} {
         constexpr auto msbOfFirstLane = T{1} << (NBits - 1);
-        this->m_v = Base::loadBaseTypeIntoLanes(values, [](auto x) { return x ? msbOfFirstLane : 0; });
+        this->m_v = Base::loadIntoLanes(values, [](auto x) { return x ? msbOfFirstLane : 0; });
     }
 
     // Booleanness is stored in the MSBs
--- a/inc/zoo/swar/SWAR.h
+++ b/inc/zoo/swar/SWAR.h
@@ -77,7 +77,8 @@ struct SWAR {
                 ~(~T(0) << NBits),
         // Use LowerBits in favor of ~MostSignificantBit to not pollute
         // "don't care" bits when non-power-of-two bit lane sizes are supported
-        LowerBits = MostSignificantBit - LeastSignificantBit;
+        LowerBits = MostSignificantBit - LeastSignificantBit,
+        MaxUnsignedLaneValue = LeastSignificantLaneMask;
 
     template <typename Arg, std::size_t N, typename = std::enable_if_t<N == Lanes, int>>
     constexpr
@@ -89,7 +90,6 @@ struct SWAR {
         m_v = result;
     }
 
-    constexpr static T MaxUnsignedLaneValue = ~(((~T{0}) << (NBits - 1)) << 1);
 
     SWAR() = default;
     constexpr explicit SWAR(T v): m_v(v) {}
--- a/inc/zoo/swar/SWAR.h
+++ b/inc/zoo/swar/SWAR.h
@@ -96,12 +96,19 @@ struct SWAR {
 
     template <typename U>
     constexpr static auto from_array(const U (&values)[Lanes]) noexcept {
-        return SWAR{from_range(std::begin(values), std::end(values))};
+        using std::begin; using std::end;
+        return SWAR{from_range(begin(values), end(values))};
+    }
+
+    template <typename U>
+    constexpr static auto from_array(const std::array<T, Lanes> &values) noexcept {
+        using std::begin; using std::end;
+        return SWAR{from_range(begin(values), end(values))};
     }
 
     using ArrayType = std::array<T, Lanes>;
 
-    constexpr SWAR(const ArrayType& array) : m_v{from_range(array.begin(), array.end())} {}
+    constexpr SWAR(const ArrayType &array) : m_v{from_range(array.begin(), array.end())} {}
 
     template <typename Arg, std::size_t N, typename = std::enable_if_t<N == Lanes, int>>
     constexpr
@@ -119,6 +126,28 @@ struct SWAR {
     SWAR() = default;
     constexpr explicit SWAR(T v): m_v(v) {}
     constexpr explicit operator T() const noexcept { return m_v; }
+
+//     constexpr auto operator==(T (&values)[Lanes]) const noexcept {
+//         return compareToContainer(values);
+//     }
+//
+//     constexpr auto operator==(std::array<T, Lanes> values) const noexcept {
+//         return compareToContainer(values);
+//     }
+
+    template <typename B>
+    constexpr bool compareToContainer(B b) const noexcept {
+        auto a = to_array();
+        if (a.size() != b.size()) {
+            return false;
+        }
+        for (auto i = 0; i < Lanes; ++i) {
+            if (a[i] != b[i]) {
+                return false;
+            }
+        }
+        return true;
+    }
 
     constexpr T value() const noexcept { return m_v; }
 
@@ -209,6 +238,9 @@ struct SWAR {
 
 template <int NBits, typename T, typename Arg>
 SWAR(Literals_t<NBits, T>, const Arg (&values)[SWAR<NBits, T>::Lanes]) -> SWAR<NBits, T>;
+
+template <int NBits, typename T>
+SWAR(Literals_t<NBits, T>, const std::array<T, SWAR<NBits, T>::Lanes>&) -> SWAR<NBits, T>;
 
 /// Defining operator== on base SWAR types is entirely too error prone. Force a verbose invocation.
 template<int NBits, typename T = uint64_t>
--- a/inc/zoo/swar/SWAR.h
+++ b/inc/zoo/swar/SWAR.h
@@ -81,12 +81,11 @@ struct SWAR {
         LowerBits = MostSignificantBit - LeastSignificantBit,
         MaxUnsignedLaneValue = LeastSignificantLaneMask;
 
-    template <typename U, typename ManipulationFn>
-    constexpr auto loadIntoLanes(const U (&values)[Lanes], const ManipulationFn&& manipulation) {
+    template <typename U>
+    constexpr auto loadIntoLanes(const U (&values)[Lanes]) const noexcept {
         auto result = T{0};
         for (auto value : values) {
-            auto laneValue = manipulation(value);
-            result = (result << NBits) | laneValue;
+            result = (result << NBits) | value;
         }
         return result;
     }
@@ -94,7 +93,7 @@ struct SWAR {
     template <typename Arg, std::size_t N, typename = std::enable_if_t<N == Lanes, int>>
     constexpr
     SWAR(Literals_t<NBits, T>, const Arg (&values)[N])
-      : m_v{loadIntoLanes(values, [](auto x) { return x; })} {}
+      : m_v{loadIntoLanes(values)} {}
 
     SWAR() = default;
     constexpr explicit SWAR(T v): m_v(v) {}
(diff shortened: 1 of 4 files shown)


inc/zoo/swar/math.h
```diff
@@ -0,0 +1,75 @@
+#pragma once
+#include "SWAR.h"
+#include <cstddef>
+#include <cstdint>
+
+namespace zoo::math {
```
thecppzoo, line 6: You know, should you have used `maths`, I would have let it slide.  But since you didn't, if you try to use the British `maths` I will reject the renaming on the basis that there is no need to make a change like that
;-)

inc/zoo/swar/math.h
```diff
@@ -0,0 +1,75 @@
...
+
+template <typename IntegerType = size_t>
+constexpr static
+std::enable_if_t<std::is_integral_v<IntegerType>, bool>
+is_power_of_two(IntegerType x) noexcept {
```
thecppzoo, line 11: Why `snake_case` instead of the normal `camelCase` and `DromedaryCamelCase` customary in the rest of `zoo` for these things?

inc/zoo/swar/math.h
```diff
@@ -0,0 +1,75 @@
+#pragma once
+#include "SWAR.h"
+#include <cstddef>
+#include <cstdint>
```
thecppzoo, line 4: `SWAR.h` includes this, or its equivalent, `stdint.h`, I see no point in things like `std::uint64_t` instead of just `uint64_t`

inc/zoo/swar/math.h
```diff
@@ -0,0 +1,75 @@
...
+
+namespace zoo::math {
+
+template <typename IntegerType = size_t>
+constexpr static
+std::enable_if_t<std::is_integral_v<IntegerType>, bool>
+is_power_of_two(IntegerType x) noexcept {
+  return x > 0 && (x & (x - 1)) == 0;
```
thecppzoo, line 12: Indentation.
Why the `X > 0` instead of the customary `X` or `bool(X)`?

inc/zoo/swar/math.h
jamierpond: habit, let's move to camelCase to be consistent. will start email thread about case to ask if you have strong preference to avoid polluting PR with unrelated discussion.

test/swar/BasicOperations.cpp
```diff
@@ -93,6 +96,53 @@ static_assert(BS{Literals<4, u16>, {F, T, F, F}}.value() == 0b0000'1000'0000'000
...
+                     S{S::Literal,      {1, 2, 3, 4}}).value()
+                     == BS{BS::Literal, {T, T, T, T}}.value());
+
+static_assert(equals(S{S::Literal,      {1, 2, 3, 4}},
+                     S{S::Literal,      {5, 6, 7, 8}}).value()
+                     == BS{BS::Literal, {F, F, F, F}}.value());
+
+static_assert(equals(S{S::Literal,      {1, 2, 3, 4}},
+                     S{S::Literal,      {5, 2, 7, 4}}).value()
+                     == BS{BS::Literal, {F, T, F, T}}.value());
```
thecppzoo, line 114: NO ACCIDENTAL INDENTATION LEVELS

thecppzoo:
I have repeatedly expressed that accidental indentation levels are contrary to the principles of how the sources ought to express more detail.
Besides, they have the practical effect of shifting the important part of the source code to the right, and an incentive to damage the code with wider lines.

inc/zoo/swar/SWAR.h
```diff
@@ -63,6 +66,7 @@ constexpr std::make_unsigned_t<T> lsbIndex(T v) noexcept {
 template<int NBits_, typename T = uint64_t>
 struct SWAR {
     using type = std::make_unsigned_t<T>;
+    using Boolean = BooleanSWAR<NBits_, T>;
```
thecppzoo, line 69: I don't like this forward reference.
It is not that intuition tells me that MSVC won't like this and give us subtle compilation problems because of MSVCs bugs, but something deeper.  I don't think that the abstraction of `BooleanSWAR` is in any way something that `SWAR` ought to advertise.  There's something fishy here that I can not express well.

inc/zoo/swar/math.h
```diff
@@ -0,0 +1,69 @@
...
+std::enable_if_t<
+    std::is_integral_v<IntegerType> &&
+    isPowerOfTwo<size_t, N>(), size_t>
+moduloPowerOfTwo(IntegerType x) noexcept {
+    return x & (N - 1);
+}
+
+static_assert(moduloPowerOfTwo<4>(0) == 0);
+static_assert(moduloPowerOfTwo<8>(9) == 1);
+static_assert(moduloPowerOfTwo<4096>(4097) == 1);
```
thecppzoo, line 33: I know we have some pollution like this in our headers.  It is pollution we haven't bother to clean.  Please move these legitimate compile time tests to their "translation unit" out of the header.

inc/zoo/swar/math.h
```diff
@@ -0,0 +1,69 @@
...
+    return x_minus_1;
+}
+// todo subtract K unsafe using BitmaskMaker
+// todo subtract K "saturated" using BitmaskMaker
+
+template <typename S>
+constexpr static auto isPowerOfTwo(S x) noexcept {
+    constexpr auto NBits = S::NBits;
+    using T = typename S::type;
+    auto greater_than_0 = greaterEqual_MSB_off(x, S{0});
```
thecppzoo, line 51: Does this work if the MSB is not off?
I think that, for example, a SWAR of 4 bits per lane with one lane containing the MSB, like for example 8, will fail.
I'm sorry that the `_MSB_off` is confusing

inc/zoo/swar/math.h
```diff
@@ -0,0 +1,69 @@
...
+template <size_t N, typename S>
+constexpr static
+std::enable_if_t<zoo::math::isPowerOfTwo<size_t, N>(), S>
+moduloPowerOfTwo(const S x) noexcept {
+    constexpr auto N_minus_1 = N - 1;
+    constexpr auto N_in_lanes = zoo::meta::BitmaskMaker<typename S::type, N_minus_1, S::NBits>::value;
+    auto y = x.value() & N_in_lanes;
+    return S{y};
+}
+
```
thecppzoo, line 66: The way this is formulated is perhaps not useful:
For integral types, when the compiler can prove the modulus is a power of two, the compiler will, for sure, use bitwise conjunction rather than the assembler for the modulo instruction.
There remains the question of when the compiler doesn't know whether the modulus is a power of two, this would be only at runtime, and then, programmers have explicitly accounted for this improvement since ages, I don't think this way of going about it will give them an incentive to use this primitive.
This is useful for "weird" types that represent integers, like multi-precision integers, I don't know, that do not satisfy the "Integral" property required by `isPowerOfTwo`.
There is one aspect that **I really like** aobut this code: the structural guarantee that you won't pass a non-power-of-two as template argument, I see myself making some code with lane sizes not powers of two, and in some remote part of the code there is an assumption broken of the lane size being a powr of two, invoking bitwise conjuction, for "hilarity".

test/swar/BasicOperations.cpp
```diff
@@ -1,11 +1,14 @@
+#include "zoo/swar/SWAR.h"
```
thecppzoo, line 1: Why?

test/swar/BasicOperations.cpp
jamierpond: this is not accidental, but just not intentionally formatted yet. didn't feel super obvious to me what this should look like.

inc/zoo/swar/SWAR.h
jamierpond: Yeah I wasn't sure about this either... wanted to see what you thought about this.

inc/zoo/swar/math.h
jamierpond: Good callout, thanks.

test/swar/BasicOperations.cpp
jamierpond: probably an auto-import

test/swar/BasicOperations.cpp
```diff
@@ -86,31 +86,29 @@ ARRAY_TEST(S8_64, 255, 255, 255, 255, 255, 255, 255, 255);
...
-
-static_assert(equals(S{S::Literal,      {1, 2, 3, 4}},
-                     S{S::Literal,      {5, 2, 7, 4}}).value()
-                     == BS{BS::Literal, {F, T, F, T}}.value());
+template <typename S = S, typename BS = BS>
+constexpr auto laneWiseEqualsTest(
+        const typename S::type (&left)[S::Lanes],
+        const typename S::type (&right)[S::Lanes],
+        const bool (&expected)[S::Lanes]) {
+    return  equals(S{S::Literal, left}, S{S::Literal, right}).value()
```
thecppzoo, line 106: Accidental levels of indentation, scopes are not closed at the indentation level.
Do you think moving the end of the argument list to the right helps the code in any way?
Why is the declaration of the three arguments more nested than the implementation?

test/swar/BasicOperations.cpp
```diff
@@ -86,31 +86,29 @@ ARRAY_TEST(S8_64, 255, 255, 255, 255, 255, 255, 255, 255);
...
-
-static_assert(equals(S{S::Literal,      {1, 2, 3, 4}},
-                     S{S::Literal,      {5, 6, 7, 8}}).value()
-                     == BS{BS::Literal, {F, F, F, F}}.value());
-
-static_assert(equals(S{S::Literal,      {1, 2, 3, 4}},
-                     S{S::Literal,      {5, 2, 7, 4}}).value()
-                     == BS{BS::Literal, {F, T, F, T}}.value());
+template <typename S = S, typename BS = BS>
+constexpr auto laneWiseEqualsTest(
```
thecppzoo, line 102: This is not right, if you want "all lanes equal" you use the underlying value comparison as in all other tests.
Unit tests ought to demonstrate how to use a codebase.  Using per-lane comparisons for a whole-base-type comparison is inadequate.
There is a probem with `equals` which is that is terseness invites incorrect use.
Even if you use `equals`, the comparison against a `BooleanSWAR` of all `true` is inelegant:
`not(boolean(~equals(`, since we have the conversion to boolean

test/swar/BasicOperations.cpp
```diff
@@ -86,31 +86,29 @@ ARRAY_TEST(S8_64, 255, 255, 255, 255, 255, 255, 255, 255);
...
+        const typename S::type (&left)[S::Lanes],
+        const typename S::type (&right)[S::Lanes],
+        const bool (&expected)[S::Lanes]) {
+    return  equals(S{S::Literal, left}, S{S::Literal, right}).value()
+            == BS{BS::Literal, expected}.value();
+}
+static_assert(laneWiseEqualsTest({1, 2, 3, 4}, {1, 2, 3, 4}, {1, 1, 1, 1}));
+static_assert(laneWiseEqualsTest({1, 2, 3, 4}, {5, 6, 7, 8}, {0, 0, 0, 0}));
+static_assert(laneWiseEqualsTest({1, 2, 3, 4}, {5, 2, 7, 4}, {0, 1, 0, 1}));
```
thecppzoo, line 111: This is legitimate as written

test/swar/BasicOperations.cpp
```diff
@@ -86,31 +86,29 @@ ARRAY_TEST(S8_64, 255, 255, 255, 255, 255, 255, 255, 255);
...
+template <typename S = S, typename BS = BS>
+constexpr auto laneWiseEqualsTest(
+        const typename S::type (&left)[S::Lanes],
+        const typename S::type (&right)[S::Lanes],
+        const bool (&expected)[S::Lanes]) {
+    return  equals(S{S::Literal, left}, S{S::Literal, right}).value()
+            == BS{BS::Literal, expected}.value();
+}
+static_assert(laneWiseEqualsTest({1, 2, 3, 4}, {1, 2, 3, 4}, {1, 1, 1, 1}));
+static_assert(laneWiseEqualsTest({1, 2, 3, 4}, {5, 6, 7, 8}, {0, 0, 0, 0}));
```
thecppzoo, line 110: These are whole-`BooleanSWAR` comparisons that can be achieved via the boolean conversion

test/swar/BasicOperations.cpp
thecppzoo, line 102: `laneWiseEquals` is problematic:
What does it mean? there are results that are expected to be equal or different according to a collection of expected boolean values, that is not even suggested by the name.
Then, `laneWiseEquals` suggests LANE WISE EQUALS.  I guess that you call your "expected" comparison "WISE EQUAL".
Notwithstanding other problems, perhaps `lanewiseEqualityComparisonExpected`?

test/swar/BasicOperations.cpp
```diff
@@ -121,33 +119,31 @@ static_assert(math::moduloPowerOfTwo<4096>(4097) == 1);
...
-
+template <typename S, typename BS>
+constexpr auto powerOfTwoTest(
+        const typename S::type (&input)[S::Lanes],
+        const bool (&expected)[S::Lanes]) {
+    return isPowerOfTwo(S{S::Literal, input}).value() == BS{BS::Literal, expected}.value();
+}
+static_assert(powerOfTwoTest<S, BS>({1, 2, 3, 4}, {1, 1, 0, 1}));
+static_assert(powerOfTwoTest<S, BS>({2, 3, 64, 77}, {1, 0, 1, 0}));
+static_assert(powerOfTwoTest<S, BS>({3, 65, 128, 0}, {0, 0, 1, 1}));
```
thecppzoo, line 130: This much copy&pasting is not valuable.
Did you consider packing all of the redundancy into an `X` macro?

thecppzoo (changes requested):
We need to document much better what are the primitives available and especially common idioms around them.
In this PR there is clear misuse of existing operations and the introduction of idioms that are not well justified, because the common idioms are not well known.
I've not worked in cleaning up the SWAR tests because I believe that they ought to be generated by a code generator where you express the intent and the test cases are generated and their source code.
We just have to stop making this more confusing by accidental misuse.

thecppzoo:
The implementation look good.
I have a problem with the unit tests.  Not even with what is tested, that seems right, but how things are tested, however semantically correct, they are misuse of the available operations.

inc/zoo/swar/math.h
Scottbruceheart: I wouldn't have let 'maths' slide.

PR #88, opened 2024-06-01
Clean up of benchmarks for string functions

thecppzoo:
1. Refactoring of the files that compose the string function benchmarks:
    1. Promotion of some utilities to the main library
2. Refactoring of the functions to reflect better their similarity as offered in the symposium:
    1. Identification of patterns and expression into templates
3. Extension of benchmarks to see the influence of parameters such as number of leading spaces for `atoi`, `atol`


benchmark/atoi.cpp
```diff
@@ -54,7 +54,7 @@ uint32_t calculateBase10(zoo::swar::SWAR<8, uint64_t> convertedToIntegers) noexc
     return uint32_t(by10001base2to32.value() >> 32);
 }
 
-uint64_t calculateBase10(zoo::swar::SWAR<8, __uint128_t> convertedToIntegers) noexcept {
+uint64_t calculateBase10_128(zoo::swar::SWAR<8, __uint128_t> convertedToIntegers) noexcept {
```
jamierpond, line 57: The naming`calculateBase10` vs `calculateBase10_128` is possibly confusing from an API perspective given they both ostensibly take in a 128bit SWAR. 

It wouldn't be obvious as a user of these functions which one to choose.

Is there also a world in which we would make this function a part of an API and the users can choose which implementation to use from a template? 

```cpp
calculateBase10<UseMaxNativeRegisterSize>(...); // default 
calculateBase10<Emulate128Bits>(...);
```

benchmark/atoi.cpp
```diff
@@ -69,6 +69,19 @@ uint64_t calculateBase10(zoo::swar::SWAR<8, __uint128_t> convertedToIntegers) no
...
+    using S = zoo::swar::SWAR<8, uint64_t>;
+    uint64_t
+        high = v >> 64,
+        low = (v << 64) >> 64;
+    auto
+        highNumbers = calculateBase10(S{low}),
+        lowNumbers = calculateBase10(S{high});
+    auto combine = highNumbers * __uint128_t(100'000'000) + lowNumbers;
+    return combine;
+}
```
jamierpond, line 83: this reads very nicely.

benchmark/atoi.cpp
```diff
@@ -194,7 +207,14 @@ auto PowersOf10Array() {
     return rv;
 };
 
-template<typename Return>
+
+template<
+    typename Return,
+    std::make_unsigned_t<Return>
+        (*CB10)(
```
jamierpond, line 214: For a _brief_ moment I was confused at `CB10`. "Is this a star wars robot?" "Callback 10?" no... "Convert to Base 10?"... _ah yes probably..._

I felt a little daft for that not clicking immediately. But, it did get me thinking. 

I think in the context of this specific file, as it is today, this name is probably fine. However, `CB10` could be clearer since it's part of the API of the function.

People may not be viewing the function in the context of the whole file. They may see the functionIntelliSense or whatever language server they are using, and all users will see is `CB10`.

Perhaps this won't be super clear to users of this function. 

Anyway. This got me thinking about the obfuscation as a signifier for (un-)importance. This comment kind of started becoming a whole email about those thoughts, so I'll send that to you in an email. 

Wow, this comment sure did get side-tracked.

Anyway, I'm sure this thought gives us something to discuss tomorrow (today!).

benchmark/atoi.cpp
```diff
@@ -256,13 +276,18 @@ Return c_strToIntegral(const char *str) noexcept {
 }
 
 int c_strToI(const char *str) noexcept {
-    return impl::c_strToIntegral<int>(str);
+    return impl::c_strToIntegral<int, calculateBase10>(str);
```
jamierpond, line 279: From reading this line of code, it gave me a thought and I'd like to see if you guys think it has any merit. 

Does it make sense to allow people to provide signed types in the interface to SWARs and the like? The signedness has semantic meaning which we disregard since we only really are caring about the total numbers of bits available in that type. 

Would it be meaningful to SFINAE the assertion of unsigned types to the API of relevant places to effectively communicate that to users? 

Not sure if huge value in this, but an idea I wanted to see if you guys thought there was any merit to the idea.

benchmark/atoi.h
```diff
@@ -19,21 +19,15 @@ std::size_t c_strLength(const char *s);
...
 
-inline int compareAtoi(const char *s) {
-    auto
-        from_stdlib = atoi(s),
-        from_zoo = c_strToI(s);
-    if(from_stdlib != from_zoo) { throw 0; }
-    return from_stdlib;
-}
-
+template<long long (*FUN)(const char *s)>
```
jamierpond, line 24: i love this. 

why not extend the abstaction? 

```cpp
template<long long (*FUN_A)(const char *s), long long (*FUN_B)(const char *s)>
inline int compareStringConsumer(const char *s) {
    auto
        a = FUN_A(s),
        b = FUN_B(s);
    if(a != b) {
        auto a = FUN_A(s);
        throw 0; }
    return from_stdlib;
}
```

jamierpond: i'm sure you could generalise it even more than that to be fair. you could make this compare any single arg functions where the return type has operator== implemented. 

```cpp
auto result = compareFunctions<strlen, zoo_strlen>(input);
```

jamierpond (commented):
Overall looks great. I've written some slightly rambling ideas. and I hope they are some combination of useful and/or entertaining!

benchmark/atoi-catch2-tests.cpp
```diff
@@ -51,3 +52,8 @@ TEST_CASE("Atol", "[pure-test][swar][atoi]") {
         REQUIRE(converted == expected);
     }
 }
+
+TEST_CASE("String function regressions", "[pure-test][atoi][swar]") {
+    auto input = "\x80\0";
+    REQUIRE(1 == zoo::c_strLength(input));
+}
```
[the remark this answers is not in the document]
thecppzoo: This is a regression function: something that early in development was a bug.  I had a broken method that would allow specifically the `char 0x80` to be mis-identified as a null.
Regressions are ad-hoc.  They don't have to be systematic.
If enough to resolve, please 'resolve conversation'

benchmark/atoi.cpp
thecppzoo: These are not, at the moment, user-facing APIs.  This is changing.

benchmark/atoi.cpp
thecppzoo: This is just the name of a function pointer parameter.  I'm not to blame for the confusing syntax of function pointers.  This is not meant to be user interface either.
I just needed to factor the calculation function out of the body of the conversion function.
There is more work, starting with not assuming the conversion is to base 10.

benchmark/atoi.cpp
thecppzoo: The reason we need the base type to be unsigned is the need for shift rights to be logical and not arithmetic:
In arithmetic shift rights ("signed" shift right) the most significant bit is copied down.  From the SWAR point of view, this behavior affects only the most significant lane.
I will create an issue to assign it to you to implement Lane-wise arithmetic shift right.
I have been clear about this from the beginning, after Robin Hood, when we bgan working again on the library, it is when I decided that we can not really support signed base types, and thus we may as well let users specify a signed type as a base to use its unsigned form.  The deficiency of this approach is that `SWAR<8, int>` and `SWAR<8, unsigned>` are different types.
We need to separate the user-facing `SWAR` from the implementation, but I am still not sure of how to do this; I mean I do not have all the answers clear to just do it.

benchmark/atoi.h
thecppzoo: Yes, I will generalize as requested, leave this conversation "unresolved"

benchmark/c_str-functions/c_str-impl.h
```diff
@@ -0,0 +1,24 @@
+#include "zoo/swar/SWAR.h"
+
+/// \brief Best implementation of Lemire's technique
+///
+/// 128 integers are currently an emulation that uses 64 computation
```
Scottbruceheart, line 5: '128 bit integers are currently an emulation that uses 64 bit integer hardware' reads better

benchmark/c_str-functions/c_str-impl.h
```diff
@@ -0,0 +1,24 @@
...
+/// together with arquitectural instructions such as
```
Scottbruceheart, line 6: typo

benchmark/c_str-functions/c_str-impl.h
```diff
@@ -0,0 +1,24 @@
...
+/// 1. Add with carry
+/// 2. Shifts widened to two general purpose register source/destination operands
+/// 3. Multiplication widened.
+/// In general, 128 bit processing would have significant penalties.
```
Scottbruceheart, line 10: 'would have' -> 'does have'

benchmark/c_str-functions/c_str-impl.h
```diff
@@ -0,0 +1,24 @@
+#include "zoo/swar/SWAR.h"
+
+/// \brief Best implementation of Lemire's technique
```
Scottbruceheart, line 3: Mention what Lemire's technique is, lest the reader of the comments not know

thecppzoo:
The latest b's in my Intel Mac:
```
--------------------------------------------------------------------------------------------------------
Benchmark                                                              Time             CPU   Iterations
--------------------------------------------------------------------------------------------------------
runBenchmark<CorpusLeadingSpaces, InvokeGLIB_Spaces>               37117 ns        37114 ns        18780
runBenchmark<CorpusLeadingSpaces, InvokeZooSpaces>                 20646 ns        20643 ns        34561
runBenchmark<CorpusDecimalDigits<8>, InvokeLemire>                  1856 ns         1856 ns       363986
runBenchmark<CorpusDecimalDigits<8>, InvokeZoo>                     1801 ns         1801 ns       391029
runBenchmark<CorpusDecimalDigits<8>, InvokeLIBC>                   26249 ns        26246 ns        26517
runBenchmark<CorpusDecimalDigits<16>, InvokeZOO_16_SPLIT>           2513 ns         2512 ns       274329
runBenchmark<CorpusDecimalDigits<16>, InvokeZOO_16_UINT128_T>       3350 ns         3350 ns       198527
runBenchmark<CorpusDecimalDigits<16>, InvokeComparison1>           39005 ns        38998 ns        17927
runBenchmark<CorpusDecimalDigits<16>, InvokeComparison2>           39013 ns        39005 ns        17759
runBenchmark<CorpusStringLength, InvokeLIBC_STRLEN>                 5352 ns         5350 ns       133537
runBenchmark<CorpusStringLength, InvokeZOO_STRLEN>                  6694 ns         6692 ns        98551
runBenchmark<CorpusStringLength, InvokeZOO_NATURAL_STRLEN>          6183 ns         6182 ns       109368
runBenchmark<CorpusStringLength, InvokeGENERIC_GLIBC_STRLEN>        9872 ns         9871 ns        66770
runBenchmark<CorpusStringLength, InvokeZOO_AVX>                     2452 ns         2451 ns       285460
runBenchmark<CorpusAtoi, InvokeGLIBC_atoi>                         35931 ns        35926 ns        20277
runBenchmark<CorpusAtoi, InvokeZOO_c_strToI>                       17988 ns        17983 ns        41016
runBenchmark<CorpusAtoi, InvokeZOO_c_strToL>                       25555 ns        25551 ns        27668
runBenchmark<CorpusAtoi, InvokeZOO_c_strToL_uint128_t>             27019 ns        27014 ns        26428
runBenchmark<CorpusAtoi, InvokeGLIBC_ATOL>                         32694 ns        32663 ns        21315
runBenchmark<CorpusAtoi, InvokeCOMPARE_ATOI>                       51678 ns        51675 ns        13833
runBenchmark<CorpusAtoi, InvokeCOMPARE_ATOL>                       61771 ns        61723 ns        10833
runBenchmark<CorpusStringLength, RepeatZooStrlen>                   6715 ns         6715 ns        96629
runCompressions<4, UseSWAR>                                         8094 ns         8093 ns        87543
runCompressions<4, UseBuiltin>                                     11444 ns        11440 ns        60482
runCompressions<4, CompareBuiltinAndSWAR>                          19028 ns        19027 ns        36711
runCompressions<8, UseSWAR>                                        15051 ns        15047 ns        43405
runCompressions<8, UseBuiltin>                                      4731 ns         4731 ns       146050
runCompressions<8, CompareBuiltinAndSWAR>                          20156 ns        20150 ns        34703
runCompressions<16, UseSWAR>                                       24827 ns        24825 ns        28390
runCompressions<16, UseBuiltin>                                     2671 ns         2671 ns       260879
runCompressions<16, CompareBuiltinAndSWAR>                         27768 ns        27766 ns        24924
runCompressions<32, UseSWAR>                                       35616 ns        35596 ns        19175
runCompressions<32, UseBuiltin>                                     1689 ns         1689 ns       421908
runCompressions<32, CompareBuiltinAndSWAR>                         37574 ns        37564 ns        18709
runCompressions<64, UseSWAR>                                       50670 ns        50669 ns        13486
runCompressions<64, UseBuiltin>                                      307 ns          306 ns      2250283
runCompressions<64, CompareBuiltinAndSWAR>                         69014 ns        68997 ns        10265
```

benchmark/c_str-functions/c_str.h
```diff
@@ -17,23 +18,18 @@ std::size_t leadingSpacesCount(const char *) noexcept;
...
-    if(from_stdlib != from_zoo) { throw 0; }
-    return from_stdlib;
-}
-
-inline int compareAtol(const char *s) {
-    auto
+template<int64_t (*FUN)(const char *s), int64_t (*BASE)(const char *) = atoll>
+inline int64_t compareAtol(const char *s) {
+    int64_t
         from_stdlib = atoll(s),
```
jamierpond, line 29: should this not be `from_stdlib = BASE(s),` given your change to the template?

thecppzoo: Very important catch!

jamierpond: With cpp 20 we can pass lambdas and fps much easier with auto templating

PR #89, opened 2024-06-02
First pass of encoding 'find first zero byte of a string' as a generic operation in zoo SWAR. Also updates equals() to be cheaper and tested.

Scottbruceheart:
The description of 'leftmost' and 'rightmost' byte gets very confusing with little or big endian representations.

Also swaps the impl of equals/differents with the `h&~(x|((x|h)-l))` where `x = l^r` construction from TAOCP, which passes tests.

### DIFF
--- a/benchmark/atoi.cpp
+++ b/benchmark/atoi.cpp
@@ -305,6 +305,8 @@ std::size_t c_strLength(const char *s) {
         // subtracting one (if their value is greater than 0x80).
         // This provides a way to detect the first null: It is the first lane
         // in firstNullTurnsOnMSB that "flipped on" its MSB
+        // According to The Art of Computer Programming, Knuth 4A pg 152,
+        // credit due to Alan Mycroft
         auto cheapestInversionOfMSBs = ~bytes;
         auto firstMSBsOnIsFirstNull =
             firstNullTurnsOnMSB & cheapestInversionOfMSBs;
--- a/inc/zoo/swar/SWAR.h
+++ b/inc/zoo/swar/SWAR.h
@@ -83,8 +83,12 @@ struct SWAR {
         PaddingBitsCount = BitWidth % NBits,
         SignificantBitsCount = BitWidth - PaddingBitsCount,
         AllOnes = ~std::make_unsigned_t<T>{0} >> PaddingBitsCount, // Also constructed in RobinHood utils: possible bug?
-        LeastSignificantBit = meta::BitmaskMaker<T, std::make_unsigned_t<T>{1}, NBits>::value,
-        MostSignificantBit = LeastSignificantBit << (NBits - 1),
+        LeastSignificantBit = meta::BitmaskMaker<T, std::make_unsigned_t<T>{1ull}, NBits>::value,
+        // Simply shifting over Least causes problems with lanes that don't fit the SWAR exactly.
+        MostSignificantBit = meta::BitmaskMaker<T, std::make_unsigned_t<T>{1ull<<(NBits - 1)}, NBits>::value,
+        // Computing LeastSignificantLaneMask with uint16_t results in an
+        // unknown narrowing or type conversion that causes use of .at() and
+        // similar with 16bit sized SWARS to fail.
         LeastSignificantLaneMask =
             sizeof(T) * 8 == NBits ? // needed to avoid shifting all bits
                 ~T(0) :
@@ -287,6 +291,7 @@ struct BooleanSWAR: SWAR<NBits, T> {
     // Booleanness is stored in the MSBs
     static constexpr auto MaskMSB =
         broadcast<NBits, T>(Base(T(1) << (NBits -1)));
+    static constexpr auto AllTrue = MaskMSB;
     static constexpr auto MaskLSB =
          broadcast<NBits, T>(Base(T(1)));
     // Turns off LSB of each lane
@@ -332,6 +337,10 @@ struct BooleanSWAR: SWAR<NBits, T> {
     constexpr BooleanSWAR(Base initializer) noexcept:
         SWAR<NBits, T>(initializer)
     {}
+
+    template<int NB, typename TT>
+    friend constexpr auto
+    equals(SWAR<NB, TT>, SWAR<NB, TT>) noexcept;
 
     template<int N, int NB, typename TT>
     friend constexpr BooleanSWAR<NB, TT>
@@ -472,13 +481,59 @@ booleans(SWAR<NB, T> arg) noexcept {
 template<int NBits, typename T>
 constexpr auto
 differents(SWAR<NBits, T> a1, SWAR<NBits, T> a2) {
-    return booleans(a1 ^ a2);
+    return ~equals(a1, a2);
 }
 
+/**
+ * @return BooleanSWAR that contains a true value in each lane where the two
+ * input SWARs match.
+ */ 
 template<int NBits, typename T>
 constexpr auto
-equals(SWAR<NBits, T> a1, SWAR<NBits, T> a2) {
-    return ~differents(a1, a2);
+equals(SWAR<NBits, T> a1, SWAR<NBits, T> a2) noexcept {
+    // Knuth, TAOCP 4A pg 152
+    using S = swar::SWAR<NBits, T>;
+    constexpr auto TAOCP_H{S{S::MostSignificantBit}}, TAOCP_L{S{S::LeastSignificantBit}};
+    auto
+        nullLaneIfEqual = a1 ^ a2,
+        allowSubtractionByOneTurningOnMSB = nullLaneIfEqual | TAOCP_H,
+        nullDetectorViaFlippingMSB_low = allowSubtractionByOneTurningOnMSB - TAOCP_L,
+        equalLanesHaveMSB_off = nullLaneIfEqual | nullDetectorViaFlippingMSB_low,
+        flipMSB = ~equalLanesHaveMSB_off,
+        rv = TAOCP_H & flipMSB;
+    return rv;
+}
+
+/**
+ * \brief finds the first null (least significant lane zero) using less compute than finding the first in greaterEqual(S{0}, x) or equals(x, S{0})
+ * @return psuedoBooleanSWAR. If there are no null lanes present, will be all
+ * zeros.  If there is a null lane present, that lane will be all 1s, and any
+ * less significant lane will be all zeros. Anything more significant than the
+ * least significant null lane will be unspecified.
+ */
+template<int NBits, typename T>
+constexpr auto
+firstZeroLane(SWAR<NBits, T> x) {
+    // Knuth, TAOCP 4A pg 152
+    using S = swar::SWAR<NBits, T>;
+    const auto h = S{S::MostSignificantBit}, l = S{S::LeastSignificantBit};
+    return h & ( x - l) & ~x;
+}
+
+/**
+ * \brief finds the first matching lane (least significant) using less compute than calling equals(a1, a2)
+ * As firstZeroLane, except we test against a given swar, lanewise.
+ * Less operations than equals(), but costs 1 more xor than firstZeroLane.
+ * @return psuedoBooleanSWAR. No matches: all zero. Matching lane: all 1s, and
+ * any less significant lane will be all zeros. Anything more significant will
+ * be unspecified.
+ */ 
+template<int NBits, typename T>
+constexpr auto
+firstMatchingLane(SWAR<NBits, T> a1, SWAR<NBits, T> a2) {
+    using S = swar::SWAR<NBits, T>;
+    const auto h = S{S::MostSignificantBit}, l = S{S::LeastSignificantBit};
+    return h & ( (a1^a2) - l) & ~(a1^a2);
 }
 
 /*
--- a/test/swar/BasicOperations.cpp
+++ b/test/swar/BasicOperations.cpp
@@ -2,6 +2,7 @@
 
 #include "catch2/catch.hpp"
 
+#include <string.h>
 #include <type_traits>
 
 using namespace zoo;
@@ -10,6 +11,9 @@ using namespace zoo::swar;
 using S2_64 = SWAR<2, uint64_t>;
 using S2_32 = SWAR<2, uint32_t>;
 using S2_16 = SWAR<2, uint16_t>;
+
+using S3_16 = SWAR<3, uint16_t>;
+using S3_32 = SWAR<4, uint32_t>;
 
 using S4_64 = SWAR<4, uint64_t>;
 using S4_32 = SWAR<4, uint32_t>;
@@ -469,7 +473,7 @@ TEST_CASE(
     "greaterEqualMSBOn",
     "[swar][unsigned-swar]"
 ) {
-    SECTION("single") {
+    SECTION("single S2_16") {
         for (uint32_t i = 1; i < 4; i++) {
             const auto left = S2_16{0}.blitElement(1,  i);
             const auto right = S2_16{S2_16::AllOnes}.blitElement(1, i-1);
@@ -477,7 +481,7 @@ TEST_CASE(
             CHECK(test.value() == greaterEqual<2, u16>(left, right).value());
         }
     }
-    SECTION("single") {
+    SECTION("single S4_32") {
         for (uint32_t i = 1; i < 15; i++) {
             const auto large = S4_32{0}.blitElement(1,  i+1);
             const auto small = S4_32{S4_32::AllOnes}.blitElement(1, i-1);
@@ -485,7 +489,7 @@ TEST_CASE(
             CHECK(test.value() == greaterEqual<4, u32>(large, small).value());
         }
     }
-    SECTION("allLanes") {
+    SECTION("allLanes S4_32") {
         for (uint32_t i = 1; i < 15; i++) {
             const auto small = S4_32(S4_32::LeastSignificantBit * (i-1));
             const auto large = S4_32(S4_32::LeastSignificantBit * (i+1));
@@ -493,6 +497,20 @@ TEST_CASE(
             CHECK(test.value() == greaterEqual<4, u32>(large, small).value());
         }
     }
+    // This uncovered an extension required for non-power of two lane sizes for
+    // equals/greaterEqual/etc related to the broadword nullcheck technique
+    // from taocp pg 152
+    // TODO(sbruce) figure out a performant fix.
+/*
+    SECTION("single 3_16") {
+        for (uint32_t i = 1; i < 3; i++) {
+            const auto left = S3_16{0}.blitElement(1,  i);
+            const auto right = S3_16{S3_16::AllOnes}.blitElement(1, i-1);
+            const auto test = S3_16{0}.blitElement(1, 2);
+            CHECK(test.value() == greaterEqual<3, u16>(left, right).value()); 
+        }
+    }
+*/
 }
 
 static_assert(0x123 == SWAR<4, uint32_t>(0x173).blitElement(1, 2).value());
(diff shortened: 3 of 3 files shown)


inc/zoo/swar/SWAR.h
```diff
@@ -430,13 +430,49 @@ booleans(SWAR<NB, T> arg) noexcept {
...
 constexpr auto
 differents(SWAR<NBits, T> a1, SWAR<NBits, T> a2) {
-    return booleans(a1 ^ a2);
+    return ~equals(a1, a2);
 }
 
+/*
+ * Returns a BooleanSWAR that contains a true value in each lane where the two
+ * inputs SWARs match.
+ */
```
thecppzoo, line 439: Nice.
If it is the same to you, please use Doxygen/Autodoc format of `/// \return or /** \return or /// @return` etc.
There are several tools, including XCode that automatically parse those comments

test/swar/BasicOperations.cpp
```diff
@@ -5,6 +5,7 @@
 #include <ios>
 #include <iomanip>
 #include <iostream>
+#include <cstring>
```
thecppzoo, line 8: I don't see value in using the `std::` instead of the functions in the global namespace.
Why `ctstring`?

inc/zoo/swar/SWAR.h
```diff
@@ -430,13 +430,49 @@ booleans(SWAR<NB, T> arg) noexcept {
...
+ * inputs SWARs match.
+ */ 
 template<int NBits, typename T>
 constexpr auto
 equals(SWAR<NBits, T> a1, SWAR<NBits, T> a2) {
-    return ~differents(a1, a2);
+    using S = swar::SWAR<NBits, T>;
+    const auto h = S::MostSignificantBit, l = S::LeastSignificantBit;
+    const auto x = (a1 ^ a2).value();
+    return BooleanSWAR<NBits, T>(h & ~(x| ((x|h) -1)));
```
thecppzoo, line 446: This code seems to have a mistake: subtracting the number 1 instead of `l` (small caps L).  You would not have had that problem if you would have used SWAR instead of unsigned operators (`x` is an unsigned, not a SWAR).

I don't doubt that this is correct (substituting 1 by `l`), but this code does not attempt to explain what is happening.
See the other implementations of SWAR operations:
1. There is an effort to explain the intermediate results
2. In the more modern implementations, the effort intensified, for example by giving a name to each partial result or each stage of the technique.

Regardless that probably `x` is what TAOCP names this, we can say `auto differentLanes = a1 ^ a2;`
In the same way we do it all over the codebase, we don't say `h` but MSB or something to that effect.

inc/zoo/swar/SWAR.h
```diff
@@ -430,13 +430,49 @@ booleans(SWAR<NB, T> arg) noexcept {
...
+ * Less operations than equals()
+ */ 
+template<int NBits, typename T>
+constexpr auto
+leastSignificantMatchingLane(SWAR<NBits, T> a1, SWAR<NBits, T> a2) {
+    using S = swar::SWAR<NBits, T>;
+    const auto h = S::MostSignificantBit, l = S::LeastSignificantBit;
+    const T x = a1 ^ a2;
+    return h & ( x - l) & ~x;
 }
```
thecppzoo, line 476: I am seeing that these functions ought not to be promoted right away to the main library `SWAR.h`, but as something more experiemental:
I am not sure the "raw test value" they return is useful by itself.
About `leastSignificantZeroLane`, this returns a `BooleanSWAR` since there is a conjunction with the most significant bit, so, the comment that describes the return value is wrong (the lane won't be all ones).  I do not like the nomenclature of "least significant blabla", but "**first**", this is how it is *used* in pokerbotic, Robin Hood, and the string operations.  I have no qualm about using a nomenclature that assumes or implies little endian byte encoding.  Yes, there are operations that use "least significant index" or the like, but in general, precisely because of the actual *use* it emerged that their right names are "first".
Do you agree on this nomenclature?

test/swar/BasicOperations.cpp
```diff
@@ -455,3 +456,36 @@ TEST_CASE(
     CHECK(SWAR<4, u16>(0x0F00).value() == saturatingUnsignedAddition(SWAR<4, u16>(0x0800), SWAR<4, u16>(0x0800)).value()); 
     CHECK(S4_32(0x0F0C'F000).value() == saturatingUnsignedAddition(S4_32(0x0804'F000), S4_32(0x0808'F000)).value()); 
 }
+
+TEST_CASE(
+    "cheapNullChecking",
+    "[swar][unsigned-swar]"
```
thecppzoo, line 462: What is the meaning of the tag "unsigned-swar"?

test/swar/BasicOperations.cpp
```diff
@@ -455,3 +456,36 @@ TEST_CASE(
...
+) {
+    {
+        // The last value / lsb null lane is all these care about, but the
+        // implementation is so tied to the execution we check both.
+        CHECK(S4_32(0x0008'0080).value() == leastSignificantZeroLane<4, uint32_t>( S4_32(0xaaa0'aa0a)));
+        CHECK(1 == S4_32(leastSignificantZeroLane<4, uint32_t>( S4_32(0xaaa0'aa0a))).lsbIndex());
+        CHECK(S4_32(0x8888'8888).value() == leastSignificantZeroLane<4, uint32_t>( S4_32(0x1110'1110)));
+        CHECK(S4_32(0x8800'0000).value() == leastSignificantZeroLane<4, uint32_t>( S4_32(0x1025'3552)));
+        CHECK(S4_32(0x8880'0000).value() == leastSignificantZeroLane<4, uint32_t>( S4_32(0x1105'5352)));
```
thecppzoo, line 471: The formatting of these checks require work to verify.
Perhaps use normal indentation to facilitate the comprehension?

benchmark/atoi.cpp
```diff
@@ -305,6 +305,8 @@ std::size_t c_strLength(const char *s) {
         // subtracting one (if their value is greater than 0x80).
         // This provides a way to detect the first null: It is the first lane
         // in firstNullTurnsOnMSB that "flipped on" its MSB
+        // According to The Art of Computer Programming, Knuth 4A pg 152,
+        // credit due to Alan Mycroft
```
thecppzoo, line 309: I just noticed I did not credit GLIBC by mistake.

inc/zoo/swar/SWAR.h
Scottbruceheart: There is exactly one other place, in the new benchmarks, this is done.  Is this a new request, that the library start supporting Autodoc/Doxygen formats?  If we'd like to move the code over to that, sounds like something for a standalone PR.

test/swar/BasicOperations.cpp
Scottbruceheart: I'm indifferent, string.h is fine.

inc/zoo/swar/SWAR.h
Scottbruceheart: Nomenclature changes: sure I can do that.
`pseudoBooleanSWAR` concept or similar covers these return values.  We just need to tighten up language around them and define it somewhere.
I think post discussion yesterday and TAOCP reread you consider these worth putting in the main library now?

test/swar/BasicOperations.cpp
Scottbruceheart: I left it as I was testing with unsigneds, but you feel a lot more strongly about tags than I do.

inc/zoo/swar/SWAR.h
thecppzoo: We should do it for all new code, in a measured, reasonable way.
Like I said, if it is the same to you, you may as well use doxygen in particular.
Is there agreement?

inc/zoo/swar/SWAR.h
thecppzoo: Not the main library.
I am beginning to have anxiety about the state of `SWAR.h`.  It is an error, long term, having implemented so many operations as members of `SWAR`, either class-members or instance members.
For example, we can't have `constexpr` constants of "intrinsic types", the SIMD types that compiler makers give you via autovectorization capabilities (GCC, Clang) or the classical architecure-specific headers.
If all implementations would be free standing functions, we could specialize them or customize them.
I prefer that we don't add to `SWAR.h` things that are not mature, for what passes as mature in this codebase, this will help welcome new users.
I have gained strong conviction that your original idea is necessary, that we totally need to express these techniques and to flesh out the potential new concept of PseudoBooleanSWAR or SIMD, just not in the main header.
Reasonable?

test/swar/BasicOperations.cpp
thecppzoo: I want us to keep using tags so that eventually we converge into a good system of tagging.
I object to the name of the test.

inc/zoo/swar/SWAR.h
Scottbruceheart: Subtracting 1 is an error, I think.
How utterly strange that I ran the unit tests on my machine on this PR and they passed, giving me confidence there wasn't an error.  Could it be that our unit tests around equals() and differents() are somehow lacking? How strange.  Seems like there should be really strong tests around such a basic operation.

The intermediate results here are fairly uncomprehensible.  We're doing something with the same flavor as the xor swap temporaries technique, and most people don't have good intuition for 'ok let's smear these test values with the high/ lo boundaries of their ranges'.  We could go for a full explanation towards greater equal (hey, this is actually a strange edge case of less than zero that requires less instructions, combined with the idea that we can easily make nonzero bytes with a xor of the original test values).

I'm not even sure that's a good way of describing it!  Would some theorem prover given a more general phrasing of this operation even come up with these operands for all hardware it'll execute on? I don't know!  It seems ok to present it as working without having a good set of names for the intermediates of the operations involved.

inc/zoo/swar/SWAR.h
thecppzoo: I don't know why you say "[t]here is exactly one other place [...] this is done": this codebase, as a whole, as been using Doxygen style comments from the very beginning, even in the original Pokerbotic where the SWAR originated.

inc/zoo/swar/SWAR.h
thecppzoo: I did not write good unit tests when I was the sole contributor.  This is because of my laziness, stupidity, and because of the certainty that I gotten correct implementations obtained through other means.  You're now the victim of my unit test mistakes.
Still, please notice that if you would have used the SWAR `operator-` instead of the integer one, you'd have had a compilation error.
If you find, during normal development, that the SWAR operations get in the way of getting things done, please let me know, those are failures of ergonomics to be corrected.  If you're in a hurry, do what you have to do.
There is no doubt subtracting the *number* 1 is an error, since it affects only the least significant lane, because of the work on all the MSBs.
What do you suggest we do with preexisting unit tests?

Scottbruceheart:
Pushed some changes just for additional discussion

inc/zoo/swar/SWAR.h
thecppzoo: Ah, we can't explain what we can't understand.  Proving that the technique works is beside the point.  Just at least indicate that the terse names are because those are the names used in TAOCP

Scottbruceheart: I'll extend testing before submitting this PR now that I've had a very direct indication that there is a huge hole.

inc/zoo/swar/SWAR.h
Scottbruceheart: I'm happy to break things apart, and generally prefer smaller 'chunks' of functionality grouped together.  
I'd also prefer to break that apart in some future PR because I find I often get sidequested on trying to figure out exactly what the right breakage boundary is &etc.

inc/zoo/swar/SWAR.h
Scottbruceheart:
```enkiwa@penguin ~/src/zoo$ find . | xargs grep -i -s '@return'
xargs: unmatched single quote; by default quotes are special to xargs unless you use the -0 option
./benchmark/atoi_impl.h:/// @return a pair to indicate the aligned pointer to the base of the block
enkiwa@penguin ~/src/zoo$ find . | xargs grep -i -s '\\\\return'
xargs: unmatched single quote; by default quotes are special to xargs unless you use the -0 option```
I see exactly one place we've specified a return value cleanly, in a Doxygen understandable way. That was the point of my comment.

Sure, I can start using doxygen readable return blocks.  The point of the previous statement was "they were so absent previously I didn't even realize it was something we cared about".

Comment (thecppzoo) [reply inc/zoo/swar/SWAR.h:439]: Come on, Scott:
on `inc`,
` find . -name '*.h' -exec grep "^[[:space:]]*//[/*]" {} +`
Gives over 200 matches, and this undercounts because all of the post-documenting comments `///<` are not included.  There are many Doxygen commands.
And my original point is that if it is the same to you, please use Doxygen.

File: inc/zoo/swar/SWAR.h [hunk 8]
```diff
@@ -472,13 +481,53 @@ booleans(SWAR<NB, T> arg) noexcept {
...
-    return booleans(a1 ^ a2);
+    return ~equals(a1, a2);
 }
 
+/**
+ * @return BooleanSWAR that contains a true value in each lane where the two
+ * inputs SWARs match.
+ */ 
+template<int NBits, typename T>
+constexpr BooleanSWAR<NBits, T>
```
thecppzoo, line 492: I prefer `auto` here, especially since you explicitly build a `BooleanSWAR` at the return site.
I have no problem to leave here the explicit return type, but then return `{ expression }`.

inc/zoo/swar/SWAR.h
```diff
@@ -472,13 +481,53 @@ booleans(SWAR<NB, T> arg) noexcept {
...
+equals(SWAR<NBits, T> a1, SWAR<NBits, T> a2) noexcept {
+    // Knuth, TAOCP 4A pg 152
+    using S = swar::SWAR<NBits, T>;
+    const S x = (a1 ^ a2);
```
thecppzoo, line 496: No explicit typing, please

inc/zoo/swar/SWAR.h
```diff
@@ -472,13 +481,53 @@ booleans(SWAR<NB, T> arg) noexcept {
...
+    return BooleanSWAR<NBits, T>{S{S::MostSignificantBit} & ~(x| ((x|S{S::MostSignificantBit}) - S{S::LeastSignificantBit}))};
```
thecppzoo, line 497:
```suggestion
    constexpr auto
        TAOCP_H{S{S::MostSignificantBit}},
        TAOCP_L{S{S::LeastSignificantBit}};
    auto
           nullLaneIfEqual = a1 ^ a2,
           allowSubtractionByOneTurningOnMSB = nullLaneIfEqual | TAOCP_H,
           nullDetectorViaFlippingMSB_low = allowSubtractionByOneTurningOnMSB - TAOCP_L,
           equalLanesHaveMSB_off = x | nullDetectorViaFlippingMSB_low,
           cheapFlippingOfMSB = ~equalLanesHaveMSB_off,
           rv = TAOCP_H & cheapFlippingOfMSB;
    
    return {rv};
```

inc/zoo/swar/SWAR.h
```diff
@@ -472,13 +481,53 @@ booleans(SWAR<NB, T> arg) noexcept {
...
+ */
 template<int NBits, typename T>
 constexpr auto
-equals(SWAR<NBits, T> a1, SWAR<NBits, T> a2) {
-    return ~differents(a1, a2);
+firstZeroLane(SWAR<NBits, T> x) {
+    // Knuth, TAOCP 4A pg 152
+    using S = swar::SWAR<NBits, T>;
+    const auto h = S::MostSignificantBit, l = S::LeastSignificantBit;
+    return h & ( x.value() - l) & ~x.value();
```
thecppzoo, line 514: No!
Why doesn't this return a SWAR?
Why is this implementation not all SWAR?

inc/zoo/swar/SWAR.h
```diff
@@ -472,13 +481,53 @@ booleans(SWAR<NB, T> arg) noexcept {
...
+constexpr BooleanSWAR<NBits, T>
+equals(SWAR<NBits, T> a1, SWAR<NBits, T> a2) noexcept {
+    // Knuth, TAOCP 4A pg 152
+    using S = swar::SWAR<NBits, T>;
+    const S x = (a1 ^ a2);
+    return BooleanSWAR<NBits, T>{S{S::MostSignificantBit} & ~(x| ((x|S{S::MostSignificantBit}) - S{S::LeastSignificantBit}))};
+}
+
+/**
+ * Less operations than equals()
```
thecppzoo, line 501:
```suggestion
/** \brief detects the first null (least significant lane zero) cheaper than finding the first in `greaterEqual(S{0}, x)`
 *
```

inc/zoo/swar/SWAR.h
```diff
@@ -472,13 +481,53 @@ booleans(SWAR<NB, T> arg) noexcept {
...
+ * be unspecified.
+ */ 
+template<int NBits, typename T>
+constexpr auto
+firstMatchingLane(SWAR<NBits, T> a1, SWAR<NBits, T> a2) {
+    using S = swar::SWAR<NBits, T>;
+    const auto h = S::MostSignificantBit, l = S::LeastSignificantBit;
+    const T x = a1 ^ a2;
+    return h & ( x - l) & ~x;
 }
```
thecppzoo, line 531: Same comments as for `firstZeroLane`

inc/zoo/swar/SWAR.h
Scottbruceheart: I thought the opinion was "if we don't really know why it works explaining it isn't really useful"?

inc/zoo/swar/SWAR.h
Scottbruceheart: Is it not forced due to the usage of 'friend'?
I also prefer auto here.

test/swar/BasicOperations.cpp
```diff
@@ -565,10 +583,188 @@ TEST_CASE(
...
+    CHECK(S4_16(0x0200).value() == saturatingUnsignedAddition(S4_16(0x0100), S4_16(0x0100)).value());
+    CHECK(S4_16(0x0400).value() == saturatingUnsignedAddition(S4_16(0x0100), S4_16(0x0300)).value());
+    CHECK(S4_16(0x0B00).value() == saturatingUnsignedAddition(S4_16(0x0800), S4_16(0x0300)).value());
+    CHECK(S4_16(0x0F00).value() == saturatingUnsignedAddition(S4_16(0x0800), S4_16(0x0700)).value());
+    CHECK(S4_16(0x0F00).value() == saturatingUnsignedAddition(S4_16(0x0800), S4_16(0x0800)).value()); 
+    CHECK(S4_32(0x0F0C'F000).value() == saturatingUnsignedAddition(S4_32(0x0804'F000), S4_32(0x0808'F000)).value()); 
+}
+
+TEST_CASE(
+    "cheapNullChecking",
```
thecppzoo, line 595: 1. Let's use the feature of Catch2 of allowing test names that are not identifiers, an ugly restriction in many other frameworks including Google's
2. This is not cheap null checking, this is fist null, etc

test/swar/BasicOperations.cpp
```diff
@@ -565,10 +583,188 @@ TEST_CASE(
...
+) {
+    {
+        // The last value / lsb null lane is all these care about, but the
+        // implementation is so tied to the execution we check both.
+        CHECK(S4_32(0x0008'0080).value() == firstZeroLane<4, uint32_t>( S4_32(0xaaa0'aa0a)));
+        CHECK(1 == S4_32(firstZeroLane<4, uint32_t>( S4_32(0xaaa0'aa0a))).lsbIndex());
+        CHECK(S4_32(0x8888'8888).value() == firstZeroLane<4, uint32_t>( S4_32(0x1110'1110)));
+        CHECK(S4_32(0x8800'0000).value() == firstZeroLane<4, uint32_t>( S4_32(0x1025'3552)));
+        CHECK(S4_32(0x8880'0000).value() == firstZeroLane<4, uint32_t>( S4_32(0x1105'5352)));
+
```
thecppzoo, line 606: These checks are unacceptable, because we use the new concept of pseudo boolean SWAR, the proper checks would treat the upper bits/lanes as "don't cares".

test/swar/BasicOperations.cpp
```diff
@@ -565,10 +583,188 @@ TEST_CASE(
...
+        // firstZeroLane finds the least significant zero (little endian). This
+        // results in the least significant zero being marked, and everything more
+        // significant than it being turned into garbage data.
```
thecppzoo, line 609: Hoist this comment up

inc/zoo/swar/SWAR.h
Scottbruceheart: Returns auto, it'll SWAR as required, no?

test/swar/BasicOperations.cpp
```diff
@@ -565,10 +583,188 @@ TEST_CASE(
...
+                                                                           S4_32(0x0000'0000)));
+        CHECK(S4_32(0x0000'8000).value() == firstMatchingLane<4, uint32_t>(S4_32(0x7325'0257),
+                                                                           S4_32(0x0000'0000)));
+        CHECK(S4_32(0x0000'0000).value() == firstMatchingLane<4, uint32_t>(S4_32(0x7341'4157),
+                                                                           S4_32(0x2222'2222)));
+        CHECK(S4_32(0x0880'0800).value() == firstMatchingLane<4, uint32_t>(S4_32(0x7321'4257), 
+                                                                           S4_32(0x2222'2222)));
+        CHECK(S4_32(0x0880'0800).value() == firstMatchingLane<4, uint32_t>(S4_32(0x7321'4357), 
+                                                                           S4_32(0x2222'3333)));
+    }
```
thecppzoo, line 629: This indentation style is not acceptable.
Do not introduce accidental indentation levels.
One indentation level means one level of detail more, and nothing more.
This indentation style moves the significant code far to the right, this is counterproductive by itself.
I treat these mistakes as fundamental challenges to the doctrine represented in this codebase.  I am literally tired of fighting this every PR review, it is deplorable that you force me to repeat myself so often about this.

inc/zoo/swar/SWAR.h
Scottbruceheart: Weird artifact of some frustrating side path, thanks for catching it.

test/swar/BasicOperations.cpp
```diff
@@ -565,10 +583,188 @@ TEST_CASE(
...
+        return s + S4_16{0x0100};
+    };
+    auto incr_lane_three = [](S4_16 s, int i)  ->S4_16{
+        return s + S4_16{0x1000};
+    };
+    auto cmp = [](S4_16 test, S4_16 expected, int i) ->S4_16 {return equals(test, expected);};
+    auto expectedf_zero = [](S4_16 in, int i) ->S4_16 {return S4_16{S4_16::MostSignificantBit}.clear(0);};
+    auto expectedf_one = [](S4_16 in, int i) ->S4_16 {return S4_16{S4_16::MostSignificantBit}.clear(1);};
+    auto expectedf_two = [](S4_16 in, int i) ->S4_16 {return S4_16{S4_16::MostSignificantBit}.clear(2);};
+    auto expectedf_three = [](S4_16 in, int i) ->S4_16 {return S4_16{S4_16::MostSignificantBit}.clear(3);};
```
thecppzoo, line 654: Why the explicit typing?
This copy and paste deserves preprocessing, we gain nothing copy and pasting this except facilitating superficial reading, at the cost of missing the point of what's really going on.

inc/zoo/swar/SWAR.h
Scottbruceheart: Ok, this combined with the other type thing is a compile error side quest I got distracted from eventually fixing, thanks for noticing.
Seems like forcing a BooleanSWAR return isn't best with all auto (since that forces a mask that might be unnecessary)

thecppzoo: What unnecessary mask?
Do you mean the masking to convert a `SWAR` to a `BooleanSWAR`?
I do not think about it because in the context of a return value the optimizers will readily identify whether the masking in the conversion is redundant.

inc/zoo/swar/SWAR.h
Scottbruceheart: Might it also be useful to (have a version that) elides that final & sometimes?  The msb are still on, and we might be doing a later bit op with some other booleanSWAR anyway....

thecppzoo: I have gotten to understand this, if you see, the names explain what is going on.  If this is not understood, then I failed, but it can be complemented with a comment.  Try to understand the explanation and let me know what is missing.

test/swar/BasicOperations.cpp
Scottbruceheart: 'horrifying unit test left in because S4_16 is broken'

inc/zoo/swar/SWAR.h
thecppzoo: This returns `T`.
I don't know why you use `x.value()`

test/swar/BasicOperations.cpp
Scottbruceheart: I don't care.
I changed formatting per request "The formatting of these checks require work to verify.
Perhaps use normal indentation to facilitate the comprehension?"
```
CHECK(S4_32(0x0880'0800).value() == 
    firstMatchingLane<4, uint32_t>(S4_32(0x7321'4357), 
        S4_32(0x2222'3333)));
```
Doesn't actually help comprehension (afaict), so I figured you must have meant 'line them up so matches are obvious'.
You must have some other indentation scheme in mind to assist comprehension. Please specify what it is.

test/swar/BasicOperations.cpp
Scottbruceheart: When we codify pseudoBooleanSWAR a little better, I'll be happy to do that.
For now, I'd prefer to leave the brittle implementation specific checks in because it matches what exists and breaks if it changes for some unexpected reason.
If you really feel strongly I can post-mask, I don't have that strong of an opinion on it.

test/swar/BasicOperations.cpp
Scottbruceheart: I agree the names of tests in this file can be improved, I'll fix this one.

inc/zoo/swar/SWAR.h
Scottbruceheart:
```
const auto h = S{S::MostSignificantBit}, l = S{S::LeastSignificantBit};
```
We have to do that to ensure these computations are in SWAR type.  The constants we've got defined in SWAR I keep expecting to be SWAR themselves, but they never are, and it is annoying.  Did we have a reason for having them be T and not SWAR<NBits, T>?

thecppzoo: It is unclear which is better.
The constants, in the remote past, were used almost exclusively directly with `this->m_v`, then it was clear to keep them of the base type.  They are still implementation details.  Since they have proven to be very useful once you are working with proper SWAR types, I'm open to make them SWAR, and use `.value()` in the cases where they can't.

inc/zoo/swar/SWAR.h
Scottbruceheart: 'allowSubtraction' is a little indistinct, since "ensure MSB on so subtracts won't roll other lanes" is a bit wordy. I don't have a better candidate though. 
used 'x' instead of nullLaneIfEqual
init list return failed on compile.
Looks fine enough.

test/swar/BasicOperations.cpp
thecppzoo: I have consistently missed the opportunity to show how to use the constructs in this library of SWAR in the unit tests.  I am trying to revert that negative inertia.
Please try to illustrate the "pseudoness" in these unit tests, precisely because it is an emergent concept: if the "post-masking" becomes annoying, then perhaps we can further articulate the "pseudoness" by ergonomics for it.

inc/zoo/swar/SWAR.h
Scottbruceheart: I also strongly prefer keeping their type homogenous.
No need to change them now, I just hadn't realized I'd made the same irritating error over and over until today.
We can decide later.

test/swar/BasicOperations.cpp
thecppzoo: I do not understand the response.

test/swar/BasicOperations.cpp
Scottbruceheart: Added a small example, can improve that over time.

test/swar/BasicOperations.cpp
Scottbruceheart: Something is wrong when using .at() and similar with uint16_t underlying type. as the comment at the top of the block says, I'm leaving this test sort of obviously wretched looking to remind us that we need to figure out why uint16_t doesn't work well as a base type for SWAR.

inc/zoo/swar/SWAR.h
```diff
@@ -333,6 +338,10 @@ struct BooleanSWAR: SWAR<NBits, T> {
         SWAR<NBits, T>(initializer)
     {}
 
+    template<int NB, typename TT>
+    friend constexpr auto
+    equals(SWAR<NB, TT>, SWAR<NB, TT>) noexcept;
+
```
thecppzoo, line 344: I don't understand the meaning of declaring an `auto` friend.  Why would this be necessary?

inc/zoo/swar/SWAR.h
```diff
@@ -472,13 +481,59 @@ booleans(SWAR<NB, T> arg) noexcept {
...
+    using S = swar::SWAR<NBits, T>;
+    constexpr auto TAOCP_H{S{S::MostSignificantBit}}, TAOCP_L{S{S::LeastSignificantBit}};
+    auto
+        nullLaneIfEqual = a1 ^ a2,
+        allowSubtractionByOneTurningOnMSB = nullLaneIfEqual | TAOCP_H,
+        nullDetectorViaFlippingMSB_low = allowSubtractionByOneTurningOnMSB - TAOCP_L,
+        equalLanesHaveMSB_off = nullLaneIfEqual | nullDetectorViaFlippingMSB_low,
+        flipMSB = ~equalLanesHaveMSB_off,
+        rv = TAOCP_H & flipMSB;
+    return rv;
```
thecppzoo, line 504: This returns a `SWAR`, not a `BooleanSWAR`.

thecppzoo: The conversion to `BooleanSWAR` ought to be made in the "public", not be an implementation detail that would require being a friend of `BooleanSWAR`.
`return booleans(rv)` suffices.

test/swar/BasicOperations.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include "catch2/catch.hpp"
 
+#include <string.h>
```
thecppzoo, line 5: Why is this inclusion needed?

inc/zoo/swar/SWAR.h
Scottbruceheart: `booleans` is implemented with a call to `constantIsGreaterEqual`, so it would not suffice.
Previously, I suggested the extra op to convert from a SWAR to a BooleanSWAR could be elided in some cases by not forcing the return immediately to a boolean.

inc/zoo/swar/SWAR.h
Scottbruceheart: I believe you said you preferred the `equals` return type to be auto.

PR #93, opened 2024-06-14
Fix integer promotion caused edge with blitElement(), at(), and more!

Scottbruceheart:
Fixes leastNBitsMask, blitElement, at and more for various usage conditions. Turns out integer promotion causes things to work in a complicated way when using the same code to mess with integers of different widths.  The implemented solution feels a little hacky, but it fixes the immediate errors (and probably ones we didn't even notice). Previously, a variety of hard to predict failures would happen when using NBits counts in various constructs at 0 and sizebits(T).

The explanation is a couple fold, and at least both of these things combined with attempts to have code that handles 8,16,32 and 64 bit has resulted in various compile errors and bugs showing up over time.

Attempting to apply bitwise complement to a zero value of unsigned 8 or 16 bit integers results in pre-conversion of the 0 init unsigned value to a signed 32 bit int value.  After conversion, this becomes a -1 (and stays signed int32).

This is then given to left or right shift, causing the failure we've seen of "cannot shift a negative number".

Attempting to use ~T{1}
Given any right or left shift of any all-ones unsigned integer type T that is subject to integer promotion to signed int32 before arithmetic (including << and >> ), the following
```
(~T{0})<< any_bitcount)
(~T{0})>> any_bitcount)
```
will always result in a left operand that is negative (and thus invalid).

As an example of why "shift left 8 on u8 or 16 on u16" "works", see integer promotion explanation above.
u16{1} << 16 behaves as expected when _actually_ shifting a signed int32.  As does u16{1} << 17!!
```
src/zoo/test/swar/BasicOperations.cpp:206: FAILED:
  CHECK( 0x01u == (u16{1}<<16) )
with expansion:
  1 == 65536 (0x10000)

src/zoo/test/swar/BasicOperations.cpp:207: FAILED:
  CHECK( 0x01u == (u16{1}<<17) )
with expansion:
  1 == 131072 (0x20000)  // 17 shift on a 16 bit type worked just fine! Promotion!

src/zoo/test/swar/BasicOperations.cpp:208: FAILED:
  CHECK( 0x01u == (u16{1}<<16) -1 )
with expansion:
  1 == 65535 (0xffff)
```

(see https://en.cppreference.com/w/c/language/conversion 'Integer promotions')
to wit:
"""Note: integer promotions are applied only

as part of usual arithmetic conversions (see above)
as part of default argument promotions (see above)
to the operand of the unary arithmetic operators + and -
to the operand of the unary bitwise operator ~
to both operands of the shift operators << and >>"""

### DIFF
--- a/inc/zoo/swar/SWAR.h
+++ b/inc/zoo/swar/SWAR.h
@@ -67,6 +67,21 @@ constexpr __uint128_t lsbIndex(__uint128_t v) noexcept {
 }
 #endif
 
+// This is placed ahead of other bare manipulation functions to reduce integer
+// promotion weirdness in the constant creation of SWAR and related.
+template<int NBits, typename T>
+constexpr auto leastNBitsMask() {
+    // It was integer promotion screwing this up at equal sizes all along, special casing unblocks.
+    constexpr auto type_bits = sizeof(T) * 8;
+    if constexpr(NBits < type_bits) {
+        return (T{1}<<NBits)-1;
+    } else if constexpr(type_bits < 32) {
+        return (T{1}<<NBits)-1;
+    } else if constexpr(type_bits >= 32) {
+        return ~T{0};
+    }
+}
+
 /// Core abstraction around SIMD Within A Register (SWAR).  Specifies 'lanes'
 /// of NBits width against a type T, and provides an abstraction for performing
 /// SIMD operations against that primitive type T treated as a SIMD register.
@@ -86,13 +101,7 @@ struct SWAR {
         LeastSignificantBit = meta::BitmaskMaker<T, std::make_unsigned_t<T>{1ull}, NBits>::value,
         // Simply shifting over Least causes problems with lanes that don't fit the SWAR exactly.
         MostSignificantBit = meta::BitmaskMaker<T, std::make_unsigned_t<T>{1ull<<(NBits - 1)}, NBits>::value,
-        // Computing LeastSignificantLaneMask with uint16_t results in an
-        // unknown narrowing or type conversion that causes use of .at() and
-        // similar with 16bit sized SWARS to fail.
-        LeastSignificantLaneMask =
-            sizeof(T) * 8 == NBits ? // needed to avoid shifting all bits
-                ~T(0) :
-                ~(~T(0) << NBits),
+        LeastSignificantLaneMask = leastNBitsMask<NBits, T>(),
         // Use LowerBits in favor of ~MostSignificantBit to not pollute
         // "don't care" bits when non-power-of-two bit lane sizes are supported
         LowerBits = MostSignificantBit - LeastSignificantBit,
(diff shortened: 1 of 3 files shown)


inc/zoo/swar/SWAR.h
```diff
@@ -67,6 +67,21 @@ constexpr __uint128_t lsbIndex(__uint128_t v) noexcept {
...
+constexpr auto leastNBitsMask() {
+    // It was integer promotion screwing this up at equal sizes all along, special casing unblocks.
+    constexpr auto type_bits = sizeof(T) * 8;
+    if constexpr(NBits < type_bits) {
+        return (T{1}<<NBits)-1;
+    } else if constexpr(type_bits < 32) {
+        return (T{1}<<NBits)-1;
+    } else if constexpr(type_bits >= 32) {
+        return ~T{0};
+    }
```
jamierpond, line 82: It looks like I might be missing some context about this bit but is there an inconsistency with when/how different types will "saturate" if the NBits argument is bigger than the type has bits? 

```cpp
// Saturating behaviour for larger types.
static_assert(0xFFFFFFFFu == leastNBitsMask<64, u32>());
static_assert(0xFFFFFFFFFFFFFFFFu == leastNBitsMask<65, u64>());
static_assert(0xFFFFFFFFFFFFFFFFu == leastNBitsMask<420, u64>());

// We don't get the "saturating" effect on smaller types?
■ Static assertion failed due to requirement '65535U == leastNBitsMask<17, unsigned short>()'
static_assert(0xFFFFu == leastNBitsMask<17, u16>());

// Instead, we get this which passes the test:
static_assert(0x1FFFFu == leastNBitsMask<17, u16>());

```

test/swar/BasicOperations.cpp
```diff
@@ -388,6 +414,17 @@ static_assert(8 == lsbIndex(1<<8));
...
+static_assert(S3_16{Literals<3,u16>, {5,4,3,5,1}}.value() == S3_16{Literals<3,u16>, {5,4,3,2,1}}.blitElement(1, 5).value());
+static_assert(S3_16{Literals<3,u16>, {5,4,5,2,1}}.value() == S3_16{Literals<3,u16>, {5,4,3,2,1}}.blitElement(2, 5).value());
+static_assert(S3_16{Literals<3,u16>, {5,1,3,2,1}}.value() == S3_16{Literals<3,u16>, {5,4,3,2,1}}.blitElement(3, 1).value());
+static_assert(S3_16{Literals<3,u16>, {1,4,3,2,1}}.value() == S3_16{Literals<3,u16>, {5,4,3,2,1}}.blitElement(4, 1).value());
+
+static_assert(1 == S3_16{Literals<3,u16>, {5,4,3,2,1}}.at(0));
+static_assert(2 == S3_16{Literals<3,u16>, {5,4,3,2,1}}.at(1));
+static_assert(3 == S3_16{Literals<3,u16>, {5,4,3,2,1}}.at(2));
+static_assert(4 == S3_16{Literals<3,u16>, {5,4,3,2,1}}.at(3));
+static_assert(5 == S3_16{Literals<3,u16>, {5,4,3,2,1}}.at(4));
```
jamierpond, line 427: cool B)

inc/zoo/swar/SWAR.h
jamierpond: will resolve this lol

jamierpond (approved):

PR #95, opened 2024-06-18
Switch constantIsGreaterEqual to use greaterEqual

Scottbruceheart:
There isn't an advantage to using the previous algorithm.


thecppzoo:
Please prove the claim there's no advantage to using the older mechanism, for all cases.

Scottbruceheart:
I think there is an advantage for the msb_off version, though?

Scottbruceheart:
I believe we also previously posted godbolt results for both that showed the general version of the constant compare (constantIsGreaterEqual) is more operations than the taocp impl.

PR #96, opened 2024-06-21
Fix benchmark

jamierpond:


benchmark/atoi.cpp
```diff
@@ -116,7 +115,7 @@ std::size_t leadingSpacesCountAligned(S bytes) noexcept {
     auto otherWhiteSpace = belowEqualCarriageReturn & ~belowTab;
     auto whiteSpace = space | otherWhiteSpace;
     auto notWhiteSpace = ~whiteSpace;
-    auto rv = notWhiteSpace ? notWhiteSpace.lsbIndex() : S::Lanes;
```
thecppzoo, line 119: Why is this necessary?
`notWhiteSpace` ought to be a `BooleanSWAR`, this would not compile if it wasn't in a any platform.

benchmark/atoi.cpp
```diff
@@ -326,7 +325,7 @@ std::size_t c_strLength_natural(const char *s) {
     auto bytes = adjustMisalignmentFor_strlen(initialBytes, misalignment);
     for(;;) {
         auto nulls = zoo::swar::equals(bytes, S{0});
-        if(nulls) { // there is a null!
```
thecppzoo, line 329: Definitively not.
`equals` returns a BooleanSWAR.  You've something broken in your side.

thecppzoo:
I just noticed the master branch was broken thanks to this pull request: `equals` wasn't returning `BooleanSWAR`.

Jamie: go to the root causes of problems not the proximate causes.

benchmark/swar/compress.cpp
```diff
@@ -21,6 +21,7 @@ enum ExtractionPrimitive {
     CompareBuiltinAndSWAR
 };
 
+#if ZOO_CONFIGURED_TO_USE_AVX()
```
thecppzoo, line 24: This is actually BMI2, not AVX

PR #102, opened 2024-09-12
Parallel Suffix cleanup

jamierpond:
- **wip**
- **think this is the best**

### DIFF
--- a/inc/zoo/swar/associative_iteration.h
+++ b/inc/zoo/swar/associative_iteration.h
@@ -42,42 +42,36 @@ std::ostream &operator<<(std::ostream &out, zoo::swar::SWAR<NB, B> s) {
 
 namespace zoo::swar {
 
+constexpr auto log2_of_power_of_two = [](auto power_of_two) {
+    if (power_of_two == 0) {
+        return 0;
+    }
+    if (power_of_two == 1) {
+        return 1;
+    }
+    return __builtin_ctz(power_of_two);
+};
+
 /// \note This code should be substituted by an application of "progressive" algebraic iteration
 /// \note There is also parallelPrefix (to be implemented)
-template<int NB, typename B>
-constexpr SWAR<NB, B> parallelSuffix(SWAR<NB, B> input) {
-    using S = SWAR<NB, B>;
-    auto shiftClearingMask = S{~S::MostSignificantBit};
-
-    auto
-        doubling = input,
-        result = S{0};
-
-    auto
-        bitsToXOR = NB,
-        power = 1;
-
-    for(;;) {
-        // From the perspective of "associative iteration", this is when we ask whether to "add"
-        if (1 & bitsToXOR) {
-            result = result ^ doubling;
-            doubling = doubling.shiftIntraLaneLeft(power, shiftClearingMask);
-        }
-        bitsToXOR >>= 1;
-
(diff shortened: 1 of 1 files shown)


inc/zoo/swar/associative_iteration.h
```diff
@@ -41,51 +41,26 @@ std::ostream &operator<<(std::ostream &out, zoo::swar::SWAR<NB, B> s) {
...
-        // 001...1
-        // 00001...1
-        // 000000001...1
-        shiftClearingMask =
-            shiftClearingMask &
-                S{static_cast<B>(shiftClearingMask.value() >> power)};
-        ZTE(power << 1);
+    for (;;) {
+        result = result ^ result.shiftIntraLaneLeft(power, shiftMask);
+        if (!--log2Count) { break; }
```
jamierpond, line 59: @thecppzoo note one condition in the inner loop now

inc/zoo/swar/associative_iteration.h
```diff
@@ -409,6 +422,74 @@ constexpr auto associativeOperatorIterated_regressive(
...
+static_assert(ps<uint32_t>(0b00000000'00000111) == 0b11111111'11111111'11111111'11111101);
+static_assert(ps<uint32_t>(0b00000000'00001111) == 0b00000000'00000000'00000000'00000101);
+static_assert(ps<uint32_t>(0b00000000'00011111) == 0b11111111'11111111'11111111'11110101);
+static_assert(ps<uint32_t>(0b00000000'00111111) == 0b00000000'00000000'00000000'00010101);
+static_assert(ps<uint32_t>(0b00000000'01111111) == 0b11111111'11111111'11111111'11010101);
+static_assert(ps<uint32_t>(0b00000000'11111111) == 0b00000000'00000000'00000000'01010101);
+
+template<typename S>
+constexpr auto parallel_suffix(S input) {
+    constexpr auto log2Count = S::Lanes;
```
thecppzoo: This can't be right, the parallel suffix does not depend on the number of lanes, but the number of bits in the lanes

inc/zoo/swar/associative_iteration.h
thecppzoo: This condition is broken: it does not work with a bit count of non-powers of 2, like 7

inc/zoo/swar/associative_iteration.h
jamierpond: looks like you might be reviewing an outdated version ?

thecppzoo:
This implementation might be simple enough, sure, but it can only accept lane sizes that have a power of two number of bits.
Let's review if the implementation I made is less efficient than yours.
Otherwise, the much harder challenge of supporting any arbitrary bitcount will have to decompose the number of bits into its binary representation to make the groups, and then AI would come to bear more clearly.
In simpler terms, this implementation is like multiplication when the factor is a power of two, much easier.

jamierpond:
@thecppzoo https://godbolt.org/z/TEP36fhzh

jamierpond:
hmmm... yeah i see what you mean...

jamierpond:
ok now working with non-power of two nuim bits, just needed to idiot check
myself about the log2 impl for non-powers of two:
https://godbolt.org/z/aPa6q8r8c

PR #107, opened 2025-01-13
Draft: fullMultiplication

jamierpond:
- **add main logic**
- **add even lane mask**
- **even/odd lane mask**
- **clean up a little**

### DIFF
--- a/inc/zoo/swar/SWAR.h
+++ b/inc/zoo/swar/SWAR.h
@@ -70,6 +70,8 @@ constexpr __uint128_t lsbIndex(__uint128_t v) noexcept {
 }
 #endif
 
+
+
 /// Core abstraction around SIMD Within A Register (SWAR).  Specifies 'lanes'
 /// of NBits width against a type T, and provides an abstraction for performing
 /// SIMD operations against that primitive type T treated as a SIMD register.
@@ -106,6 +108,17 @@ struct SWAR {
             result = (result << NBits) | *first;
         }
         return result;
+    }
+
+    constexpr static auto evenLaneMask() {
+        using S = SWAR<NBits, T>;
+        static_assert(0 == S::Lanes % 2, "Only even number of elements supported");
+        using D = SWAR<NBits * 2, T>;
+        return S{(D::LeastSignificantBit << S::NBits) - D::LeastSignificantBit};
+    }
+
+    constexpr static auto oddLaneMask() {
+        return SWAR<NBits, T>{static_cast<T>(~evenLaneMask().value())};
     }
 
     template <typename Range>
@@ -594,5 +607,8 @@ static_assert(
     logarithmFloor(SWAR<8>{0xFF7F3F1F0F070301ull}).value() ==
     0x0706050403020100ull
 );
+
+static_assert(SWAR<4, u16>::evenLaneMask().value() == 0b0000'1111'0000'1111);
+static_assert(SWAR<4, u16>::oddLaneMask().value()  == 0b1111'0000'1111'0000);
 
 }}
--- a/inc/zoo/swar/associative_iteration.h
+++ b/inc/zoo/swar/associative_iteration.h
@@ -543,7 +543,6 @@ fullMultiplication(SWAR<NB, T> multiplicand, SWAR<NB, T> multiplier) {
 }
 
 using S = SWAR<4, u32>;
-
 static_assert(S::oddLaneMask().value() == 0xF0F0'F0F0);
 static_assert(S::evenLaneMask().value() == 0x0F0F'0F0F);
 
@@ -560,14 +559,6 @@ static_assert(fullMultiplication(S{0x0008'0000}, S{0x0008'0000})
 
 static_assert(fullMultiplication(S{0x0001'0000}, S{0x0008'0000})
                   .overflowed.value() == 0x0000'0000);
-
-// static_assert([] {
-//     // fullMultiplication(S{0x0008'0012}, S{0x0007'0032}).result.value()
-//     ==0x0008'0034 auto r = fullMultiplication(S{0x0003'0012},
-//     S{0x0003'0032}); if (r.result.value() != 0x0009'0034) { return false; }
-//     if (r.overflow.value() != 0x0000'0000) { return false; }
-//     return true;
-// }());
 
 static_assert(fullMultiplication(S{0x0008'0012}, S{0x0007'0032})
                   .result.value() == 0x0008'0034);
--- a/inc/zoo/swar/associative_iteration.h
+++ b/inc/zoo/swar/associative_iteration.h
@@ -510,6 +510,79 @@ constexpr auto halvePrecision(SWAR<NB, T> even, SWAR<NB, T> odd) {
     return evenHalf | oddHalf;
 }
 
+
+template <int NB, typename T> struct MultiplicationResult {
+   SWAR<NB, T> result;
+   BooleanSWAR<NB, T> overflowed;
+};
+
+// static_assert([] {
+//    using D = SWAR<8, u32>;
+//    using S = SWAR<4, u32>;
+//    using H = SWAR<2, u32>;
+//    constexpr auto UpperHalfOfLanes = S::oddLaneMask().value();
+//    static_assert(UpperHalfOfLanes == 0xF0F0'F0F0);
+//    return true;
+// }());
+
+template <int NB, typename T>
+constexpr MultiplicationResult<NB, T>
+fullMultiplication(SWAR<NB, T> multiplicand, SWAR<NB, T> multiplier) {
+   using S = SWAR<NB, T>; using D = SWAR<NB * 2, T>;
+
+   auto [l_even, l_odd] = doublePrecision(multiplicand);
+   auto [r_even, r_odd] = doublePrecision(multiplier);
+   auto res_even = multiplication_OverflowUnsafe(l_even, r_even);
+   auto res_odd = multiplication_OverflowUnsafe(l_odd, r_odd);
+
+   // Into the double precision world
+   constexpr auto HalfLane = S::NBits;
+   constexpr auto UpperHalfOfLanes = SWAR<S::NBits, T>::oddLaneMask().value();
+   auto res = halvePrecision(res_even, res_odd);
+
+   auto over_even = D{(res_even.value() & UpperHalfOfLanes) >> HalfLane};
+   auto over_odd = D{(res_odd.value() & UpperHalfOfLanes) >> HalfLane};
+   auto overflow_values = halvePrecision(over_even, over_odd);
+
+   // back to normal precision world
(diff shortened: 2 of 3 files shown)


inc/zoo/swar/associative_iteration.h
```diff
@@ -503,13 +495,69 @@ constexpr auto halvePrecision(SWAR<NB, T> even, SWAR<NB, T> odd) {
...
+
+   // back to normal precision world
+   auto did_overflow = ~(zoo::swar::equals(overflow_values, S{0}));
+
+   return {res, did_overflow};
+}
+
+using S = SWAR<4, u32>;
+
+static_assert(S::oddLaneMask().value() == 0xF0F0'F0F0);
```
jamierpond, line 538: aware these tests not formatted nicely, just making a draft for visibility

inc/zoo/swar/SWAR.h
```diff
@@ -108,6 +110,17 @@ struct SWAR {
...
 
+    constexpr static auto evenLaneMask() {
+        using S = SWAR<NBits, T>;
+        static_assert(0 == S::Lanes % 2, "Only even number of elements supported");
+        using D = SWAR<NBits * 2, T>;
+        return S{(D::LeastSignificantBit << S::NBits) - D::LeastSignificantBit};
+    }
+
+    constexpr static auto oddLaneMask() {
+        return SWAR<NBits, T>{static_cast<T>(~evenLaneMask().value())};
```
thecppzoo, line 121: We must not add things that interfere with lane sizes that are not power-of-two bits.
I think we should discuss what is the convention we will use for the bits that are not a full lane.
I think it should be a user's choice, for example, "don't care", meaning the user tolerates any values in those bits, or "always-zero", that the library guarantees no operation will set those bits.

thecppzoo (commented):
Thanks for this draft.
This work surfaces several important questions:
1. Support for non-power-of-two lane sizes
2. What are we going to do with two's complement signs? perhaps this is not `fullMultiplication` but `safeMultiplication`
3. We have to implement "negation" (two's complement flipping the sign)

That being said, we are in an excellent position to also implement division as multiplication by the reciprocal, which would be useful at least for compile-time divisors.

Please merge the `auto` declarations: this does not coerce the type, but verifies that all the declarands have the same type:
```c++
auto a = initialize_a(inputsForA);
auto b = initialize_b(inputsForB);
```
In that code, the types are not coerced (very good!) but `a` and `b` may be of different types.
```c++
auto
    a = initA(iA),
    b = initB(iB);
```
We still don't coerce the types, but if `a` and `b` have different types, it is a compilation error.
This is especially useful here in the SWAR library

thecppzoo:
Perhaps we can also make "widening multiplication", that doubles the lane size.  For example, in x86-64, there are the instructions to multiply two register-size values and get a result of double the number of bits, using the "DX:AX" for the result, so, for 64 bits, it would be RDX with the upper 64 bits, and RAX with the lower, in this way, the multiplication also widens.  Ask Claude what is the name of this.

inc/zoo/swar/associative_iteration.h
```diff
@@ -503,13 +495,66 @@ constexpr auto halvePrecision(SWAR<NB, T> even, SWAR<NB, T> odd) {
...
+
+template <int NB, typename T>
+constexpr MultiplicationResult<NB, T>
+fullMultiplication(SWAR<NB, T> multiplicand, SWAR<NB, T> multiplier) {
+   using S = SWAR<NB, T>; using D = SWAR<NB * 2, T>;
+
+   auto [l_even, l_odd] = doublePrecision(multiplicand);
+   auto [r_even, r_odd] = doublePrecision(multiplier);
+   auto res_even = multiplication_OverflowUnsafe(l_even, r_even);
+   auto res_odd = multiplication_OverflowUnsafe(l_odd, r_odd);
```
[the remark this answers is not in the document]
jamierpond: also todo signed multiplication

inc/zoo/swar/associative_iteration.h
```diff
@@ -503,13 +466,94 @@ constexpr auto halvePrecision(SWAR<NB, T> even, SWAR<NB, T> odd) {
...
     auto
         evenHalf = RV{even.value()} & HalvingMask,
         oddHalf = RV{(RV{odd.value()} & HalvingMask).value() << NB/2};
     return evenHalf | oddHalf;
 }
 
+
+template <int NB, typename T> struct MultiplicationResult {
+   SWAR<NB, T> result;
+   SWAR<NB, T> overflow;
```
thecppzoo, line 479: This is not overflow.

inc/zoo/swar/associative_iteration.h
```diff
@@ -503,13 +466,94 @@ constexpr auto halvePrecision(SWAR<NB, T> even, SWAR<NB, T> odd) {
...
+};
+
+template <int NB, typename T>
+constexpr auto
+doublingMultiplication(SWAR<NB, T> multiplicand, SWAR<NB, T> multiplier) {
```
thecppzoo, line 483: `doubling` is confusing here.  `doublePrecisionMultiplication` is fine, `multiplicationByDoublingPrecision`, ...

inc/zoo/swar/associative_iteration.h
```diff
@@ -503,13 +466,94 @@ constexpr auto halvePrecision(SWAR<NB, T> even, SWAR<NB, T> odd) {
...
+   auto [l_even, l_odd] = doublePrecision(multiplicand);
+   auto [r_even, r_odd] = doublePrecision(multiplier);
+   auto
+       res_even = multiplication_OverflowUnsafe(l_even, r_even),
+       res_odd = multiplication_OverflowUnsafe(l_odd, r_odd);
+   return SWAR_Pair<NB * 2, T>{res_even, res_odd};
+}
+
+template <int NB, typename T>
+constexpr MultiplicationResult<NB, T>
```
thecppzoo, line 494: Why the explicit return type?

inc/zoo/swar/associative_iteration.h
```diff
@@ -503,13 +466,94 @@ constexpr auto halvePrecision(SWAR<NB, T> even, SWAR<NB, T> odd) {
...
+    constexpr auto NumBitsPerLane = S::NBits;
+    constexpr auto
+        MSB = S{S::MostSignificantBit},
+        LSB = S{S::LeastSignificantBit};
+
+    auto operation = [](auto left, auto right, auto counts) {
+      auto mask = makeLaneMaskFromMSB(counts);
+      auto product = saturatedMultiplication(left, right);
+      return (product & mask) | (left & ~mask);
+    };
```
thecppzoo, line 538: Please, let us not repeat already-identified deficiencies!

inc/zoo/swar/associative_iteration.h
```diff
@@ -503,13 +466,94 @@ constexpr auto halvePrecision(SWAR<NB, T> even, SWAR<NB, T> odd) {
...
+   auto lane_mask = did_overflow.MSBtoLaneMask();
+   auto saturated = result | lane_mask;
+   return S{saturated};
+}
+
+
+// TODO(Jamie): Add tests from other PR.
+template<int NB, typename T>
+constexpr auto saturatingExponentiation(
+    SWAR<NB, T> x,
```
thecppzoo, line 525: Absolutely not.
We're not removing the non-saturating exponentiation and provide only the saturating exponentiation.  Don't do that.
**Always** the general operation is pre-requisite for the more specific.

inc/zoo/swar/associative_iteration.h
```diff
@@ -474,8 +474,8 @@ constexpr auto halvePrecision(SWAR<NB, T> even, SWAR<NB, T> odd) {
 
 
 template <int NB, typename T> struct MultiplicationResult {
-   SWAR<NB, T> result;
-   SWAR<NB, T> overflow;
+   SWAR<NB, T> lower;
+   SWAR<NB, T> upper;
```
thecppzoo, line 478: merge

inc/zoo/swar/associative_iteration.h
```diff
@@ -497,11 +497,11 @@ wideningMultiplication(SWAR<NB, T> multiplicand, SWAR<NB, T> multiplier) {
...
        UpperHalfOfLanes = SWAR<S::NBits, T>::oddLaneMask().value();
-   auto [res_even, res_odd] = doublingMultiplication(multiplicand, multiplier);
-   auto result = halvePrecision(res_even, res_odd);
+   auto [lower, upper] = doublingMultiplication(multiplicand, multiplier);
+   auto result = halvePrecision(lower, upper);
    auto
-       over_even = D{(res_even.value() & UpperHalfOfLanes) >> HalfLane},
-       over_odd = D{(res_odd.value() & UpperHalfOfLanes) >> HalfLane};
+       over_even = D{(lower.value() & UpperHalfOfLanes) >> HalfLane},
+       over_odd = D{(upper.value() & UpperHalfOfLanes) >> HalfLane};
```
thecppzoo, line 504: shift intra lane allows you to provide the mask.
Please use those primitives instead of deploying the pick-axe

inc/zoo/swar/SWAR.h
```diff
@@ -258,6 +258,12 @@ constexpr auto horizontalEquality(SWAR<NBits, T> left, SWAR<NBits, T> right) {
     return left.m_v == right.m_v;
 }
 
+template <int NBits, typename T>
+constexpr static auto consumeMSB(SWAR<NBits, T> s) noexcept {
+    using S = SWAR<NBits, T>;
+    auto msbCleared = s & ~S{S::MostSignificantBit};
+    return S{static_cast<T>(msbCleared.value() << 1)};
+}
```
thecppzoo, line 267: I am not sold on promoting this to the main header of swar.
This really seems to be an artifact of the "regressive" direction of "associative iteration", it does not cohere enough to the SWAR library itself.

inc/zoo/swar/associative_iteration.h
```diff
@@ -479,15 +479,16 @@ template <int NB, typename T> struct MultiplicationResult {
...
-constexpr auto
-doublingMultiplication(SWAR<NB, T> multiplicand, SWAR<NB, T> multiplier) {
-   using S = SWAR<NB, T>; using D = SWAR<NB * 2, T>;
-   auto [l_even, l_odd] = doublePrecision(multiplicand);
-   auto [r_even, r_odd] = doublePrecision(multiplier);
+constexpr
+auto
+doublePrecisionMultiplication(SWAR<NB, T> multiplicand, SWAR<NB, T> multiplier) {
+   auto
+       icand = doublePrecision(multiplicand),
```
thecppzoo, line 486: Nice! never thought about omitting the prefix

thecppzoo:
I can not resist to comment about how elegant this is all looking.
The primitives of doubling/halving precision were a success.

inc/zoo/swar/associative_iteration.h
jamierpond: good call

PR #110, opened 2025-04-29
Copy on Write 

jamierpond:
- **Comments**
- **string io affordance working**
- **works**
- **update copy**
- **copy on write tests pass**

### DIFF
--- a/test/inc/demo/type_erasure_shared_pointer_value_manager.hpp
+++ b/test/inc/demo/type_erasure_shared_pointer_value_manager.hpp
@@ -38,6 +38,7 @@ struct UserValueManagement {
         /// Abbreviation
         using SPM = SharedPointerManager;
 
+        struct ExclusiveAware {};
         // not part of the end-user interface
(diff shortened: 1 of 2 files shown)


test/inc/demo/type_erasure_shared_pointer_value_manager.hpp
```diff
@@ -57,27 +70,41 @@ struct UserValueManagement {
...
             AffordanceSpecifications::template Operation<SPM>...
         };
 
+        auto isExclusive() const noexcept {
+            auto sp = const_cast<SPM *>(this)->sharedPointer();
+            return 1 == sp->use_count();
+        }
+
+        auto makeExclusive() {
+            *sharedPointer() = std::make_shared<V>(*value());
```
patrickroberts, line 88: Not that this would be an issue for the vast majority of cases, but maybe const qualify the lvalue reference so it always selects the copy constructor.
```suggestion
            *sharedPointer() = std::make_shared<V>(static_cast<const V &>(*value()));
```

